Issue #3, opened 2016-01-27
Set up CI and tests


snowleopard:
- [x] Enable Travis CI
- [x] Add `.travis.yml`
- [x] Use `-Wall -fwarn-tabs` to keep the codebase clean
- [x] Add basic test infrastructure
- [x] ~~Enforce code formatting style.~~

mvdan:
What about `hlint`? For now all it reports is this:

```
Circuit/Concept.hs:21:28: Warning: Use .
Found:
  foldr (.||.) (const False) . map notEqual
Why not:
  foldr ((.||.) . notEqual) (const False)

Circuit/Concept.hs:32:33: Warning: Use .
Found:
  foldr (.&&.) (const True) (map after causes)
Why not:
  foldr ((.&&.) . after) (const True) causes

Circuit/Concept.hs:38:33: Warning: Use .
Found:
  foldr (.||.) (const False) (map after causes)
Why not:
  foldr ((.||.) . after) (const False) causes

3 suggestions
```

Perhaps you find it too pedantic, but I've learned a thing or two about writing better haskell with it.

snowleopard:
`hlint` is very useful but I don't always take its suggestions. I guess we wouldn't want to enforce it.

However, we might integrate it with PRs anyway so that it gives some suggestions to consider. Is there a way to do that so that it doesn't influence the CI status?

mvdan:
Sure, but if its warnings are not fatal, CI would be green and people wouldn't go check what's going on. And doing so would be slow too.

So perhaps it's just better for each one to run it after writing new pieces of code, if not confident that the code will already be warning-free.

snowleopard:
Agreed!

mvdan:
Can you update the progress? Only the first tests are left.

mvdan:
Looked into code formatting, but not convinced. Mostly because there isn't an agreed-upon or vastly popular style to use.

snowleopard:
Done. Thanks @mvdan!

mvdan:
We cannot use hindent yet, see https://github.com/chrisdone/hindent/issues/170

snowleopard:
OK, I'll close this now. We can come back to the code style issue once we find the right tool for this.

PR #6, opened 2016-01-28
Add first test, simply testing a buffer

mvdan:
Updates #3.

### DIFF
--- a/Tests.hs
+++ b/Tests.hs
@@ -1,0 +1,34 @@
+import System.Exit (exitFailure)
+import Text.Printf
+
+import Circuit
+import Circuit.Concept
+import Circuit.Dynamics
+
+main = do
+    testBuffer
+
+assertEq want got
+    | want /= got = do
+        printf "--- FAIL:\nwant: "
+        print want
+        printf "got:  "
+        print got
+        exitFailure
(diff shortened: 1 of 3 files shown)


Tests.hs
```diff
@@ -0,0 +1,34 @@
...
+    | otherwise = do
+        printf "OK "
+        print want
+
+data Signal = A | B deriving (Eq, Show, Enum, Bounded)
+
+testBuffer = do
+    putStrLn "=== testBuffer"
+    assertEq True $ initial circuit initState
+    assertEq [Transition A True] $ enabledTransitions initState circuit
```
snowleopard: I think the following would be a bit more readable?

``` haskell
assertEq [rise A] $ enabledTransitions initState circuit
let afterA = fire (rise A) initState
assertEq [fall A, rise B] $ enabledTransitions afterA circuit
```

snowleopard:
@mvdan Looks good, just one comment above.

Tests.hs
mvdan: Done, thanks for the suggestion! I knew it looked way too verbose.

snowleopard:
:+1:

PR #16, opened 2016-02-03
CI: Show details when running tests

mvdan:
This is important to be able to verify that the tests are indeed being
run as part of CI. If none of them are run and the output is hidden, we
might not notice at all.


mvdan:
This is how it looks now. If you think the "OK" lines are too verbose, we can remove them. I initially used them to verify that everything was working properly, but they might be too much once we have more tests.

PR #18, opened 2016-02-13
Take some sane hlint suggestions

mvdan:


Test.hs
```diff
@@ -45,7 +44,7 @@ shouldBe :: (Eq a, Show a) => a -> a -> Simulation Signal IO ()
 shouldBe x y = lift $ assertEq x y
 
 shouldReturn :: (Eq a, Show a) => Simulation Signal IO a -> a -> Simulation Signal IO ()
-shouldReturn x y = (flip shouldBe y) =<< x
+shouldReturn x y = (`shouldBe` y) =<< x
```
snowleopard: Nice. Maybe we can take this further and swap the order:

``` haskell
x >>= (`shouldBe` y)
```

mvdan: SGTM, will update

PR #22, opened 2016-03-03
appveyor.yml: no need to download cabal

mvdan:


snowleopard:
Awesome. Thanks for cleaning up this mess!

PR #23, opened 2016-03-22
Add program to simulate at runtime

mvdan:
You'll need `hint` installed.

Usage:

```
 $ runghc Simulate.hs concept-examples/Concept.hs
Signal names: ["A","B","C"]
Circuit signals: 3

State: 000
Enabled: [A+,B+]
Do: A+
State: 100
Enabled: [B+]
Do: B+
State: 110
Enabled: [C+]
Do: C+
State: 111
Enabled: [A-,B-]
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,7 @@
 # Concepts
 
-[![Build Status](https://travis-ci.org/tuura/concepts.svg?branch=master)](https://travis-ci.org/tuura/concepts) [![Build status](https://ci.appveyor.com/api/projects/status/dn6igqdxf3cq2t8w/branch/master?svg=true)](https://ci.appveyor.com/project/snowleopard/concepts/branch/master)
+[![Build Status](https://travis-ci.org/tuura/concepts.svg?branch=master)](https://travis-ci.org/tuura/concepts)
+[![Build status](https://ci.appveyor.com/api/projects/status/dn6igqdxf3cq2t8w/branch/master?svg=true)](https://ci.appveyor.com/project/snowleopard/concepts/branch/master)
 
 A DSL for asynchronous circuits specification.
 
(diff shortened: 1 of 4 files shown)


snowleopard:
This is great. Ping once you are happy with the result.

mvdan:
IMO the only thing holding this from merging is the hard-coding of `Signal`. So once you get the signal stuff worked out in master, I can probably adapt both the sample input file and `Main.hs` to use it, allowing for `concept-simulate` to work with any input and not just those using `A | B | C`.

snowleopard:
OK, I'll do my best to finish the `Signal` stuff soon.

mvdan:
@snowleopard I removed the WIP as the two things left to do are done.

There are two major pitfalls with the current implementation that I left as TODOs. The first is that currently `A | B | C` don't play any role, other than giving the names to `[0..n-1]`. The other is that we assume that the initial state is in terms of `a` currently, so it will probably break if it's something other than `const Bool`.

I've documented everything, but it's not easy to read and understand. So we can go through it together tomorrow if you like.

mvdan:
Also, I've left all commits as they illustrate the history of how I came up with this result. I'll squash before we merge.

mvdan:
What do you think about making the inital state in terms of arguments too? For example:

```
initialState :: (Eq a) => a -> a -> a -> State a
initialState a b c = State (== a)
```

Or perhaps the initial state shouldn't be defined in the user-provided file, along with the circuit? I could ask for it to be specified interactively at the start of the program.

mvdan:
@snowleopard stack (windows) is using an older version of hint (`0.4.3`, not `0.5.1`). Any idea why that would be? cabal (linux) is using the correct version (the latest).

snowleopard:
:+1: Windows build is failing though -- can we fix that?

snowleopard:
:tada:

Issue #17, opened 2016-02-03
Input, output and internal events and signals


snowleopard:
Right now we don't distinguish between input/output/internal events and signals. I think this is one of the reasons we can't quite figure out the right way to deal with initialisation concepts #11.

Also, STGs do require to assign signals into groups so we will have to do it anyway at some point. Let's use this issue to discuss this.

snowleopard:
A temporary solution is in the `signal-types` branch, in particular, see `src/Tuura/Concept/Signal.hs`.

However, it is not satisfactory at the moment, as one can mix `Input a` and `Output a` in one specification and it will be allowed. We probably don't want to permit this, which seems to lead to a complete separation of input/internal/output types. 

Does it make sense to consider concepts parameterised by `i o t` instead of simply `a`? This looks overly verbose, inconvenient and inextensible, but is simpler (does not require GADTs or existential types) and solves the problem of mixing signals with the same name but different types in a single specification.

snowleopard:
Note, STGs are usually defined on a set of signals `S` that is partitioned into non-overlapping sets `I`, `O`, and `T` corresponding to input, output and internal signals, respectively. We need to encode this partition nicely, preferably at the type level, and make sure we don't allow creating signals of different types but with the same name (my previous solution fails in this aspect).

snowleopard:
We could treat `Input a` and `Output a` as completely different signals, which seems to solve the problem, but may be confusing.

jrbeaumont:
It seems we are stuck between a rock and a hard place.

`Input a` and `Output a` could be confusing, but it's a much nicer way of viewing it. It'd be better this way I feel however. 

Could we not add some verification for non-overlap of signals? It would mean the nice implementation can be used, and avoid some problems.

snowleopard:
There is one reason we might want both `Input a` and `Output a` to be allowed. Imagine specifying behaviour of two circuits connected with a common wire:

``` haskell
data Signal = A | B | C

-- For circuit1 signal B is an output
circuit1 = rise (Input A) ~> rise (Output B) <> ...

-- For circuit2 signal B is an input
circuit2 = rise (Input B) ~> rise (Output C) <> ...

-- In the overall system, we compose the circuits and B is either internal or output
system = circuit1 <> circuit2 <> ...
```

So, we have the following possible interplay of signal types above:
- `circuit1` has input `A` and output `B`
- `circuit2` has input `B` and output `C`
[...]

PR #24, opened 2016-03-23
WIP: Signal types

snowleopard:
This is a work in progress on type-safe signals.

### DIFF
--- a/signal-types-hacking.hs
+++ b/signal-types-hacking.hs
@@ -1,42 +1,52 @@
 {-# LANGUAGE DataKinds, GADTs, TypeFamilies #-}
 {-# OPTIONS_GHC -fno-warn-unticked-promoted-constructors #-}
+{-# OPTIONS_GHC -fno-warn-name-shadowing #-}
 
-data SignalType = I | O
+-- Internal (I), output (O) and internal (T) signal types
+data SignalTag = I | O | T
 
 data Signal t a where
-    Input  :: a -> Signal I a
-    Output :: a -> Signal O a
+    Input    :: a -> Signal I a
+    Output   :: a -> Signal O a
+    Internal :: a -> Signal T a
 
-type InputSignal  = Signal I
-type OutputSignal = Signal O
+fromSignal :: Signal t a -> a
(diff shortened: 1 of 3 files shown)


mvdan:
To make it easier to differentiate between PRs which are ready to merge and those which aren't, you usually prefix the title with `WIP:` like on mine :)

If I understand this correctly, `a` can be of any type?

snowleopard:
> WIP

Done! :)

> If I understand this correctly, `a` can be of any type?

Yes, you are right. I do not put any constraints on the underlying representation of signals. However, in some cases we do ask for `Eq a`, `Show a` and other instances to be able to compare, show signals, etc.

mvdan:
> Yes, you are right. I do not put any constraints on the underlying representation of signals. However, in some cases we do ask for Eq a, Show a and other instances to be able to compare, show signals, etc.

Could we use this to somehow limit these to singletons of some kind? Like you could define `"a"` as a signal once, but not twice.

snowleopard:
Each _signal name_ like `name = "a"` can be defined at most once in Haskell, but there is no way to limit the number of times it can be used in other expressions. So, it may be used as part of both `Input name` and `Output name`. I don't think we can easily forbid this unfortunately.

mvdan:
What if we simply iterated over the signals defined by the user and checked that the values made up a set, i.e. no repetitions?

You can argue this won't show up at compile time. But in the end we'd be compiling at runtime, so this would be a check that would run right after the compiling.

snowleopard:
Well, the whole point of going through `GADTs` is to provide signal type checking at compile time. If we give it up, we can simply use `data Signal a = Input a | Output a` constructor that will work fine for runtime checks. /cc #17

mvdan:
What should we do with this? I'm inclined to think that we should close it after the idea you had with wrappers. Circuits should take a number of arguments, and then the wrapping functions will let users easily make use of them with their own signal types.

snowleopard:
Agreed, let me close this. I'll probably keep the branch for a while, it's an interesting experiment.

PR #27, opened 2016-04-05
WIP: Wrappers

snowleopard:
Experimenting with circuit wrappers, see #25.


jrbeaumont:
@snowleopard Can we close this PR now? I believe we've implemented this now.

snowleopard:
I don't think we've implemented wrappers yet (correct me if I'm wrong!), so let's keep this open. It's very far behind the master branch, but maybe useful to keep until we come back to implementing wrappers properly.

jrbeaumont:
You are correct, I apologise, I was thinking it was something to do with signal types.

PR #28, opened 2016-05-17
Simplify simulate, cleanup after first MR

mvdan:


mvdan:
Also, the behaviour with the bundled example is exactly the same as before.

snowleopard:
:+1: Merged.

PR #39, opened 2016-08-24
Defined initial states will now be correctly translated.

jrbeaumont:
Initial states will no longer be hard-coded as `False`.

I also included a new function to allow defining several of the same initial state for a list of signals at once, `initials`, for ease of use. 

All examples have been updated to conform to the requirements of the tool:
- All signals now must have their initial state defined. 
- The buck scenarios now use `initials` for clarity. 
- C-element examples continue to use the original `initialise` to show options when defining initial states.

### DIFF
--- a/src/Tuura/Concept/Concat/Abstract.hs
+++ b/src/Tuura/Concept/Concat/Abstract.hs
@@ -1,12 +1,12 @@
 module Tuura.Concept.Concat.Abstract (
     Interface (..),
+    InitialValue (..),
     Concept (..),
     initialConcept, arcConcept,
     interfaceConcept,
-    (.&&.), (.||.),
     ) where
 
-import Control.Applicative
+import Data.Monoid
 
 -- Abstract concepts
 -- * s is the type of states
@@ -14,9 +14,21 @@ import Control.Applicative
 
 data Interface = Unused | Input | Internal | Output deriving (Ord, Eq, Show)
 
+data InitialValue = Undefined | Defined { getDefined :: Bool } | Inconsistent deriving (Eq, Show)
+
+instance Monoid InitialValue where
+    mempty = Undefined
+
+    mappend Inconsistent _ = Inconsistent
+    mappend _ Inconsistent = Inconsistent
+    mappend Undefined x = x
+    mappend x Undefined = x
+    mappend (Defined x) (Defined y) = if x == y then Defined x else Inconsistent
+
+
 data Concept s e a = Concept
                    {
-                       initial :: s -> Bool,
+                       initial :: a -> InitialValue,
                        arcs    :: [(e, e)],
                        interface :: a -> Interface
                    }
@@ -24,13 +36,13 @@ data Concept s e a = Concept
 instance Monoid (Concept s e a) where
     mempty = Concept
              {
-                 initial = const True,
+                 initial = const mempty,
                  arcs    = [],
                  interface = const Unused
              }
     mappend a b = Concept
                   {
-                      initial = initial a .&&. initial b,
+                      initial = \s -> initial a s <> initial b s,
                       arcs    = arcs a     ++  arcs b,
                       interface = \s -> interface a s `max` interface b s
                   }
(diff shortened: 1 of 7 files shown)


src/Tuura/Concept/Concat/Abstract.hs
```diff
@@ -1,51 +1,57 @@
...
                        interface :: a -> Interface
                    }
 
 instance Monoid (Concept s e a) where
     mempty = Concept
              {
-                 initial = const True,
+                 initial = const mempty,
                  arcs    = [],
                  interface = const Unused
```
snowleopard: For symmetry, I suggest to add a `Monoid` instance for `Interface`, so we can say here `const mempty` and use `<>` instead of `max` lateron. This makes this more uniform.

src/Tuura/Concept/Concat/Abstract.hs
```diff
@@ -1,51 +1,57 @@
...
+    mempty = Undefined
+
+    mappend Inconsistent _ = Inconsistent
+    mappend _ Inconsistent = Inconsistent
+    mappend Undefined x = x
+    mappend x Undefined = x
+    mappend (Defined x) (Defined y) = if x == y then Defined x else Inconsistent
+
+
 data Concept s e a = Concept
```
snowleopard, line 34: Add a note here:

``` haskell
-- Note, type parameter s is unused in this implementation and may be removed later.
```

src/Tuura/Concept/Concat/Circuit.hs
```diff
@@ -58,8 +58,11 @@ type CircuitConcept a = Concept (State a) (Transition a) a
 consistency :: CircuitConcept a
 consistency = mempty
 
-initialise :: a -> Bool -> CircuitConcept a
-initialise a = initialConcept . after . Transition a
+initialise :: Eq a => a -> Bool -> CircuitConcept a
+initialise a v = initialConcept $ \s -> if s == a then Defined v else Undefined
+
+initials :: Eq a => [a] -> Bool -> CircuitConcept a
```
snowleopard: Another way could be to have two functions `initialise0` and `initialise1` with type `Eq a => [a] -> CircuitConcept a`, so one can simply write `initialise0 [a, b, c] <> initialise1 [d]`.

jrbeaumont:
These suggestions are good ideas, I will make these changes.

translate/Main.hs
```diff
@@ -98,15 +98,21 @@ doTranslate :: MonadIO m => [DynSignal] -> Concept (State DynSignal) (Transition
...
+            let initStrs = map (\s -> (show s, (getDefined $ initial circuit s))) signs
             let arcStrs = map (\(from, to) -> (show from, show to)) (arcs circuit)
             let inputSigns = filter ((==Input) . interface circuit) signs
             let outputSigns = filter ((==Output) . interface circuit) signs
             let internalSigns = filter ((==Internal) . interface circuit) signs
             liftIO $ putStr $ genSTG inputSigns outputSigns internalSigns initStrs arcStrs
             return ()
-        UnusedSignal ss -> liftIO $ putStr $ "Error. The following signals are not declared as input, output or internal: \n"
-            ++ unlines (map show ss)
+        Invalid ss is us -> liftIO $ do
```
snowleopard: Let's use clearer names here: `ss`, `is` and `us` are too cryptic. Perhaps just reuse names from the validation function.

snowleopard:
@jrbeaumont Great, let me know when you are done with the above and I'll merge this.

src/Tuura/Concept/Concat/Abstract.hs
```diff
@@ -1,51 +1,57 @@
...
              {
-                 initial = const True,
+                 initial = const mempty,
                  arcs    = [],
                  interface = const Unused
              }
     mappend a b = Concept
                   {
-                      initial = initial a .&&. initial b,
+                      initial = \s -> initial a s <> initial b s,
```
snowleopard: I think it may be possible to simplify this to just `initial = initial a <> initial b` because there is a standard instance generation rule: `Monoid a => Monoid (b -> a)`, i.e. if we know how to compose values of type `a`, we also know how to compose functions returning type `a`.

src/Tuura/Concept/Concat/Abstract.hs
```diff
@@ -1,51 +1,58 @@
...
 -- * s is the type of states
 -- * e is the type of events
 
-data Interface = Unused | Input | Internal | Output deriving (Ord, Eq, Show)
+data Interface = Unused | Input | Output | Internal deriving (Ord, Eq, Show)
 
+instance Monoid Interface where
+    mempty = Unused
+
+    mappend x y = x `max` y
```
snowleopard: Even shorter: `mappend = max` :-)

snowleopard:
Very nice redesign of the whole thing, I like it!

jrbeaumont:
Me too. Let's hope we can continue keeping it this nice.

PR #40, opened 2016-09-05
Stack is now used for installing and with Travis

jrbeaumont:
Changed travis file to use Stack instead of Cabal.

Appveyor already used Stack.

Stack will now be used over cabal. Workcraft and the manual will be updated accordingly.


snowleopard:
Nice! :+1:

PR #43, opened 2016-09-09
Added manual and changed modules to use STG instead of concat

jrbeaumont:


doc/manual.md
```diff
@@ -0,0 +1,676 @@
...
+use an example and go through it based on lines, explaining what is
+necessary. For this example we will be using the same as example as in
+previous sections, from the examples included with the concepts repo.
+This is “*Celement_with_env_1.hs*”.
+
+Concepts file layout
+--------------------
+
+![A concepts file](http://jrbeaumont.github.io/concepts-manual/images/concepts_file_screenshot.svg)
+
```
snowleopard, line 370: I suggest to simply use a Haskell code snippet here, e.g.:

``` haskell
module Concept where

import Tuura.Concept.Concat
...
```

A screenshop from Sublime Text editor may be confusing (readers might think it's required to use concepts).

snowleopard:
Looks very nice. Just one comment above.

jrbeaumont:
@snowleopard: As you'll see I've included all the movement of modules in this.
I also need to make some changes to Workcraft for this, as it has a default emtpy concepts file using the wrong module import

snowleopard:
Great! You'll also need to generate new screenshots, as currently they still show `Concat`.

snowleopard:
@jrbeaumont Let me know when you are done with the screenshots, and I'll merge this.

jrbeaumont:
Oh right, that's a problem.

I'll do it when I get home, I'm currently on OS X and screenshotting is a pain.

snowleopard:
No problem, take your time.

jrbeaumont:
@snowleopard: The images have been updated.

snowleopard:
OK, let's now drop type signatures `Eq a => a -> ... -> CircuitConcept a` from the docs.

jrbeaumont:
@snowleopard: Type signatures dropped. All tested and working. Manual updated.

snowleopard:
@jrbeaumont Awesome! Merged now.

PR #46, opened 2016-09-12
Changed gp and gp_ack initial states in buck examples to 1.

jrbeaumont:
Removed initial states from `me` to avoid issues with these examples, where an initial state is defined more than once.

### DIFF
--- a/examples/zcAbsent_scenario.hs
+++ b/examples/zcAbsent_scenario.hs
@@ -20,7 +20,7 @@ circuit uv oc zc gp_ack gn_ack gp gn =
     gpHandshake = handshake gp gp_ack
     gnHandshake = handshake gn gn_ack
 
-    initialState = initialise0 [uv, oc, zc, gp_ack, gn_ack, gp, gn]
+    initialState = initialise0 [uv, oc, zc, gp_ack, gp] <> initialise1 [gn_ack, gn]
 
     chargeFunc = interface <> ocFunc <> ocReact <> environmentConstraint
                 <> circuitConstraint <> gpHandshake <> gnHandshake <> initialState
--- a/examples/zcAbsent_scenario.hs
+++ b/examples/zcAbsent_scenario.hs
@@ -20,9 +20,9 @@ circuit uv oc zc gp_ack gn_ack gp gn =
     gpHandshake = handshake gp gp_ack
     gnHandshake = handshake gn gn_ack
 
-    initialState = initialise0 [uv, oc, zc, gp_ack, gp] <> initialise1 [gn_ack, gn]
+    initialState = initialise0 [uv, oc, zc, gp, gp_ack] <> initialise1 [gn, gn_ack]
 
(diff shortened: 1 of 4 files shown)


examples/zcAbsent_scenario.hs
```diff
@@ -20,7 +20,7 @@ circuit uv oc zc gp_ack gn_ack gp gn =
     gpHandshake = handshake gp gp_ack
     gnHandshake = handshake gn gn_ack
 
-    initialState = initialise0 [uv, oc, zc, gp_ack, gn_ack, gp, gn]
+    initialState = initialise0 [uv, oc, zc, gp_ack, gp] <> initialise1 [gn_ack, gn]
```
snowleopard, line 24: I find it a bit odd to list `x_ack` before `x` signals :-) Let's reorder.

examples/zcAbsent_scenario.hs
```diff
@@ -20,7 +20,7 @@ circuit uv oc zc gp_ack gn_ack gp gn =
...
 
     chargeFunc = interface <> ocFunc <> ocReact <> environmentConstraint
                 <> circuitConstraint <> gpHandshake <> gnHandshake <> initialState
```
snowleopard: Let's slightly change formatting:

``` haskell
     chargeFunc = interface <> ocFunc <> ocReact <> environmentConstraint <>
                  circuitConstraint <> gpHandshake <> gnHandshake <> initialState
```

examples/zcAbsent_scenario.hs
jrbeaumont: I believe it's like this in the examples in the article . Shall I change those too?

src/Tuura/Concept/STG/Circuit.hs
```diff
@@ -96,7 +96,6 @@ handshake11 a b = handshake a b <> initialise a True <> initialise b True
 
 me :: Eq a => a -> a -> CircuitConcept a
```
snowleopard, line 98: Let's add a note: `-- TODO: Restrict the initial state so that a=b=1 is not allowed.`

examples/zcAbsent_scenario.hs
snowleopard: I think I changed them already!

snowleopard:
Looks fine, but see some comments above.

examples/zcAbsent_scenario.hs
jrbeaumont: I have found a couple of areas I want to change for uniformity.
e.g for the `zcLateScenario`

```
circuit(uv, oc, zc, gp_ack, gn_ack, gp, gn) = ...
```

snowleopard: Sure, please do. Hope you won't clash with Danil's changes.

jrbeaumont:
@snowleopard Those should now be fixed

snowleopard:
:+1:

Issue #34, opened 2016-08-08
Include or-causality in translation.


jrbeaumont:
Or-causality currently cannot be translated, meaning some gates cannot be used, such as `OR` and `AND` gates. 

Or-causality will require a new signal-transition node for the effect transition for each possible cause. 
e.g. in `a+, b+ ~> c+`, there will need to be two `c+` transitions. 

For each and-causality where `c+` is the effect e.g `x+ ~> c+`, the cause transitions will need to connect to all `c+` transition nodes.

jrbeaumont:
@snowleopard: 
Can we just ensure that I know how OR-causality works correctly?

I'd say that:

```
[a+, b+] ~||~> c+
```

Causes there to be two `c+` transitions, one that `a+`connects to  (`c+`) and one that `b+` connects to  (`c+/1`).

Then, if something comes along, with a NAND for example, which features:

```
[x-, y-] ~||~> c+
```

We then have two more `c+` transitions, one that `x-` connects to  (`c+/2`) and one that `x-` connects to  (`c+/3`).

But as part of that NAND, we need to include:

```
x+ ~> c-
y+ ~> c-
```

And both `x+` and `y+` must connect to all `c+` transitions  (`c+`, `c+/1`, `c+/2`, `c+/3`)

snowleopard:
@jrbeaumont The best way to understand is to draw 5 consistency loops for `a`, `b`, `x`, `y`, `c` and try :)

I think here's how the examples should be translated:
- `[a+, b+] ~||~> c+` leads to two read arcs: `a1 -- c+/1` and `b1 -- c+/2`.
- `[x-, y-] ~||~> c+` leads to two read arcs: `x0 -- c+/1` and `y0 -- c+/2`.
- `[a+, b+] ~||~> c+ <> [x-, y-] ~||~> c+` leads to eight read arcs: `a1 -- c+/1`, `b1 -- c+/2`, `a1 -- c+/3`, `b1 -- c+/4`, `x0 -- c+/1`, `x0 -- c+/2`, `y0 -- c+/3`, `y0 -- c+/4`. So we have some kind of a cross-product between two OR-causality concepts.

snowleopard:
@jrbeaumont I think it would be best if you try my translation in Workcraft and check that it is correct. If you come up with a better/simpler translation -- let me know.

snowleopard:
[...]

PR #48, opened 2016-09-14
Implemented OR-causality.

jrbeaumont:
The algorithm for this I will display in a blog post. For now, here is some testing information. The test files are here: 
[Or-causality testing.zip](https://github.com/tuura/concepts/files/472367/Or-causality.testing.zip)

OR-gate C = A+B:
Works as expected. Synthesis produces this circuit:
[circuit_or_gate.pdf](https://github.com/tuura/concepts/files/472381/circuit_or_gate.pdf)

AND-gate C = AB:
Also works as expected. Synthesis produces this circuit:
[circuit_and_gate.pdf](https://github.com/tuura/concepts/files/472382/circuit_and_gate.pdf)

Two OR-gates - C = (A+B)(D+E):
Synthesis produces a predictably complex circuit, but simulation of the STG reveals it to work as expected.
To set `C` high (minimum): `A+` and `D+`.  `A+` and `E+`.  `B+` and `D+`.  `B+` and `E+`.
To set `C` low: All signals low

AND and OR gate - C = (A+B)DE:
Again, synthesis produces something unrecognizable. Simulation reveals correctness:
To set `C` high (minimum): `A+`, `D+` and `E+`. `B+`, `D+` and `E+`. 
To set `C` low (minimum): `A-`, `B-` and `E-`. `A-`, `B-` and `D-`. 

A 2-input OR and a 3-input OR - C = (A+B)(D+E+F):
Synthesis doesn't produce much, but simulation is good:
To set `C` high (minimum): `A+` and `D+`. `A+` and `E+`. `A+` and `F+`. `B+` and `D+`. `B+` and `E+`. `B+` and `F+`.
To set `C` low: All signals low

Let me know if you have any other test cases you wish me to try.

### DIFF
--- a/translate/Main.hs
+++ b/translate/Main.hs
@@ -99,8 +99,8 @@ doTranslate signs circuit = do
     case validate signs circuit of
         Valid -> do
             let initStrs = map (\s -> (show s, (getDefined $ initial circuit s))) signs
-            let orStrs = manageOrs (ors circuit)
             let arcStrs = map (\(from, to) -> (show from, show to)) (arcs circuit)
+            let orStrs = manageOrs (ors circuit)
             let inputSigns = filter ((==Input) . interface circuit) signs
             let outputSigns = filter ((==Output) . interface circuit) signs
             let internalSigns = filter ((==Internal) . interface circuit) signs
@@ -115,7 +115,7 @@ doTranslate signs circuit = do
             when (undef  /= []) $ putStr $ "The following signals have undefined initial states: \n"
                                     ++ unlines (map show undef) ++ "\n"
 
-manageOrs :: [([Transition DynSignal], Transition DynSignal)] -> [(String, String)]
+manageOrs :: [([Transition DynSignal], Transition DynSignal)] -> [String]
 manageOrs ors = do
     let effect = snd (head ors)
     -- Filter for just one of the effects
@@ -126,10 +126,18 @@ manageOrs ors = do
     -- Remap these
     let mapped = mapOrs causesLists
     let n = length mapped
-    -- Create arc pairs for these
+    -- Create arc pairs for these ors
     let orArcs = arcOrs mapped effect n
-    if (remainder /= []) then orArcs ++ manageOrs remainder else orArcs
-    -- return ("11", "22")
+    -- Create read-arcs for these
+    let orStrs = concatMap transition orArcs
+    -- Add new transitions to consistency loop
+    addSymbTransition effect n ++ concatMap transition orArcs
+    --if (remainder /= []) then orArcs ++ manageOrs remainder else orArcs
+
+addSymbTransition :: Transition DynSignal -> Int -> [String]
+addSymbTransition effect n
+        | "+" `isSuffixOf` (show effect) = map (\x -> (printf "%s0 %s/%s\n" (init (show effect)) (show effect) (show x)) ++ (printf "%s/%s %s1" (show effect) (show x) (init (show effect)))) [1..n - 1]
+        | otherwise                      = map (\x -> (printf "%s1 %s/%s\n" (init (show effect)) (show effect) (show x)) ++ (printf "%s/%s %s0" (show effect) (show x) (init (show effect)))) [1..n - 1]
 
 mapOrs :: [[Transition DynSignal]] -> [[Transition DynSignal]]
 mapOrs causesLists = addOrs (head causesLists) (tail causesLists)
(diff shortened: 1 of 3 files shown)


snowleopard:
I think we should combine `arcs` and `ors` into `arcs`:

``` haskell
arcs :: [([a], a)]

arcConcept from to = mempty { arcs = [([from], to)] }
```

In other words, usual causality is a special case of OR-causality with a singleton list of causes. Pretty sure this will simplify the rest of the code too.

jrbeaumont:
Can be done. 
Something like:
When it comes to handling OR causality in the translation, filter for lists of >1 size and those are OR causality?

snowleopard:
Yes, but I don't think you will ever need to do the filtering. Think of singletons as brackets with single variables when doing expansion.

In fact I think all you need is standard Haskell function `sequence` which will magically compute all terms for you :-)

See answers here: http://stackoverflow.com/questions/4119730/cartesian-product

snowleopard:
For example, try `sequence [[1], [2, 3], [4, 5]]`.

jrbeaumont:
Yeah that would work great. I'll replace this part in my implementation and see how it goes.

jrbeaumont:
@snowleopard: I have used your suggestions and re-tested. All works well :)

translate/Main.hs
```diff
@@ -114,6 +114,36 @@ doTranslate signs circuit = do
             when (undef  /= []) $ putStr $ "The following signals have undefined initial states: \n"
                                     ++ unlines (map show undef) ++ "\n"
 
+handleArcs :: [([Transition DynSignal], Transition DynSignal)] -> [String]
+handleArcs arcLists
+    | arcLists == [] = []
+    | otherwise = do
```
snowleopard: Why are you using `do`-notation here? Looks like you can replace this by `otherwise = addSymbTransition ... where ...` dropping a lot of `let`s.

translate/Main.hs
```diff
@@ -114,6 +114,36 @@ doTranslate signs circuit = do
...
+    -- Create arc pairs for these arcs
+    let arcs = arcarcs mapped effect n
+    -- Get remaining arcs for other effects
+    let remainder = arcLists \\ causesForEffect
+    -- return [consistency loop additions] ++ [arc strings] ++ [the same result for other effects]
+    addSymbTransition effect n ++ concatMap transition arcs ++ handleArcs remainder
+
+addSymbTransition :: Transition DynSignal -> Int -> [String]
+addSymbTransition effect n
+        | "+" `isSuffixOf` (show effect) = map (\x -> (printf "%s0 %s/%s\n" (init (show effect)) (show effect) (show x)) ++ (printf "%s/%s %s1" (show effect) (show x) (init (show effect)))) [1..n - 1]
```
snowleopard: What are you trying to achieve with `"+"`isSuffixOf`(show effect)`? Is this simply `newValue effect == True` or, even shorter, `newValue effect`?

translate/Main.hs
```diff
@@ -114,6 +114,36 @@ doTranslate signs circuit = do
...
 
+handleArcs :: [([Transition DynSignal], Transition DynSignal)] -> [String]
+handleArcs arcLists
+    | arcLists == [] = []
+    | otherwise = do
+    let effect = snd (head arcLists)
+    -- Filter for just one of the effects
+    let causesForEffect = filter (\o -> snd o == effect) arcLists
+    -- Get just the list of lists of causes
+    let causesLists = map (\l -> fst l) causesForEffect
```
snowleopard: Looks like simply `map fst` :)

translate/Main.hs
```diff
@@ -114,6 +114,36 @@ doTranslate signs circuit = do
             when (undef  /= []) $ putStr $ "The following signals have undefined initial states: \n"
                                     ++ unlines (map show undef) ++ "\n"
 
+handleArcs :: [([Transition DynSignal], Transition DynSignal)] -> [String]
```
snowleopard, line 118: As a first version this is fine -- well done! But there are a few improvements that will simplify the code even further. Let's add a todo: `TODO: 1) Get rid of explicit recursion (use groupBy)? 2) Improve performance.`

translate/Main.hs
```diff
@@ -114,6 +114,36 @@ doTranslate signs circuit = do
...
+    let remainder = arcLists \\ causesForEffect
+    -- return [consistency loop additions] ++ [arc strings] ++ [the same result for other effects]
+    addSymbTransition effect n ++ concatMap transition arcs ++ handleArcs remainder
+
+addSymbTransition :: Transition DynSignal -> Int -> [String]
+addSymbTransition effect n
+        | "+" `isSuffixOf` (show effect) = map (\x -> (printf "%s0 %s/%s\n" (init (show effect)) (show effect) (show x)) ++ (printf "%s/%s %s1" (show effect) (show x) (init (show effect)))) [1..n - 1]
+        | otherwise                      = map (\x -> (printf "%s1 %s/%s\n" (init (show effect)) (show effect) (show x)) ++ (printf "%s/%s %s0" (show effect) (show x) (init (show effect)))) [1..n - 1]
+
+arcarcs :: [[Transition DynSignal]] -> Transition DynSignal -> Int -> [(String, String)]
```
snowleopard: I'm a bit confused by the name. Can you come up with a clearer one?

translate/Main.hs
snowleopard: Also, let's try to keep lines below 80 characters for readability (e.g. can't read it in github).

snowleopard (commented):
Please fix my comments, and I think we can merge this version as long as it works. But there will be further improvements to make the code more efficient and clear.

jrbeaumont:
@snowleopard: Fixes made. Still working correctly.

jrbeaumont:
@snowleopard: Hold off for 2 minutes, I forgot to include AND and OR gate concepts

jrbeaumont:
@snowleopard: OK i'm done, unless you have any further suggestions :)

jrbeaumont:
This should close #34

snowleopard:
Great, thanks!

Issue #35, opened 2016-08-08
Allow a user to define their own signal names.


jrbeaumont:
Currently, because we do not store names for signals when defining concepts, we find the number of signals in the system and name them from `A`. 

All signals are also defined as outputs, so being able to define signals as inputs, outputs and internal will reduce a step post-translation.

jrbeaumont:
Signal types is discussed in issue #17.
This issue has been renamed to simply be based on signal names.

PR #51, opened 2016-09-16
Improved `handleArcs` function.

jrbeaumont:
Removed explicit recursion.

Added `extra` to the cabal file, required for `groupSortOn`.

### DIFF
--- a/translate/Main.hs
+++ b/translate/Main.hs
@@ -29,24 +29,26 @@ main = do
 
 {- Our own Signal type. Contains the signal index, from 0 to x-1 if
  - there are x signals. -}
-data DynSignal = Signal Int deriving Eq
+data Signal = Signal Int deriving Eq
 
-instance Show DynSignal where show (Signal i) = [chr (ord 'A' + i)]
+--instance Show Signal where show (Signal i) = [chr (ord 'A' + i)]
+instance Show Signal where
+    show (Signal i)
+        | i < 26    = [chr (ord 'A' + i)]
+        | otherwise = 'S' : show i
 
-instance Ord DynSignal
+instance Ord Signal
     where
-        (<=) (Signal x) (Signal y)
-            | x <= y = True
-            | otherwise = False
+        compare (Signal x) (Signal y) = compare x y
 
 {- Temporary module to help us use any number of signals in the user's
  - circuit. Otherwise, we would be bound to a number of arguments
  - (signals) known at compile time.
- - Keep the data DynSignal line in sync with the one above! -}
+ - Keep the data Signal line in sync with the one above! -}
 signalsApply :: Int -> [String]
 signalsApply num = [
     "import Data.Char",
-    "data DynSignal = Signal Int deriving Eq",
+    "data Signal = Signal Int deriving Eq",
     "signs = [Signal i | i <- [0.." ++ show (num-1) ++ "]]",
     "apply c = c " ++ unwords ["(signs !! " ++ show i ++ ")" | i <- [0..num-1]]]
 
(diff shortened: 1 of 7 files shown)


translate/Main.hs
```diff
@@ -32,6 +33,12 @@ data DynSignal = Signal Int deriving Eq
 
 instance Show DynSignal where show (Signal i) = [chr (ord 'A' + i)]
 
+instance Ord DynSignal
+    where
+        (<=) (Signal x) (Signal y)
```
snowleopard: If you prefer `<=` instead of `compare` you should at least simplify this as follows :-)

``` haskell
instance Ord DynSignal
    where
       Signal x <= Signal y = x <= y
```

translate/Main.hs
```diff
@@ -116,17 +123,13 @@ doTranslate signs circuit = do
...
-handleArcs arcLists
-    | arcLists == [] = []
-    | otherwise = addSymbTransition effect n ++ concatMap transition arcs ++ handleArcs remainder
+handleArcs arcLists = addSymbTransition effect n ++ concatMap transition arcs
         where
             effect = snd (head arcLists)
-            causesForEffect = filter (\o -> snd o == effect) arcLists
-            causesLists = map fst causesForEffect
+            causesLists = map fst arcLists
             mapped = sequence causesLists
```
snowleopard: I think name `mapped` is not very meaningful (maybe it used to, but lost the meaning after refactoring).

translate/Main.hs
```diff
@@ -116,17 +123,13 @@ doTranslate signs circuit = do
 
 -- TODO: 1) Get rid of explicit recursion (use groupBy)? 2) Improve performance.
```
snowleopard: We can drop the `TODO`: there is no explicit recursion any more, and the code is faster (we do sorting + one pass, instead of multiple passes).

snowleopard (commented):
A couple of comments on existing code:
- I think `boolBit` is just standard `fromEnum`.
- Shall we rename `DynSignal` to just `Signal`? I think it's clearer: `data Signal = Signal Int deriving Eq`.
- This will break when there are more than 26 signals: `show (Signal i) = [chr (ord 'A' + i)]`. Think how to improve.
- This is a bit sad: `"+"`isSuffixOf`f`. Can't you convert to strings later, so you can avoid testing strings instead of working with `Transition` type?
- In `initVals`, can't you replace `foldr (\s -> (++)...` simply with `map`?
- Can you remove explicit recursion from `arcPairs` and `initVal`?
- In `output`, use `nubOrd` from `extra`, which is faster.
- Should `symbLoop` be called `consistencyLoop` as in the paper?

jrbeaumont:
Thanks for the comments, I will get to this ASAP. 
It should be possible to do some work on most of these.

> This will break when there are more than 26 signals: show (Signal i) = [chr (ord 'A' + i)]. Think how to improve.

I agree this needs improvement, but I don't believe it can be a quick fix, and it can be fixed with #35, allowing a user to define their own signal names.

snowleopard:
> I agree this needs improvement, but I don't believe it can be a quick fix

Well, there are plenty of quick fixes! For example:

``` haskell
show (Signal i)
    | i < 26    = [chr (ord 'A' + i)]
    | otherwise = 'S' : show i
```

> and it can be fixed with #35, allowing a user to define their own signal names.

True, that would be preferable.

translate/Main.hs
jrbeaumont: I actually just forgot to change this to compare, so I'm gonna use that. It just seems a little nicer to me for some reason

jrbeaumont:
``` haskell
instance Show Signal where show (Signal i)
    | i < 26    = [chr (ord 'A' + i)]
    | otherwise = 'S' : show i
```

This throws errors, it doesn't like the `|`.
I'm gonna come back to this soon. 26 signals at this point is quite a lot, but it does need some adjustment.

snowleopard:
I think formatting is wrong, perhaps this will work:

``` haskell
instance Show Signal where
    show (Signal i)
        | i < 26    = [chr (ord 'A' + i)]
        | otherwise = 'S' : show i
```

jrbeaumont:
That works, thanks.
Next issue:

> This is a bit sad: "+"isSuffixOff. Can't you convert to strings later, so you can avoid testing strings instead of working with Transition type?

Because of the numbering of transitions (e.g `c+/1`), I can't change the first argument to `transition`, `(String, String)` to `(Transition Signal, Transition Signal)`. I can change it to `(Transition Signal, String)` which would remove `"+" isSuffixOf`.

jrbeaumont:
Made these changes now. Let me know if there's more to be done :)

snowleopard:
Merged from 10,600m above.

PR #54, opened 2016-10-22
WIP: Added tests for translation.

jrbeaumont:
@snowleopard: I have added some tests, but found some issues along the way.

In order to do the testing, the translation code needs to be installed as part of the cabal build phase, therefore I have refactored some of it into `Tuura.Concept.STG.Translation`. This allows testing to occur on `.g` output and arcs of different concepts. 

First of all, if you think these tests need some changes, please let me know.

Second of all: I have attempted to change the `translate/Main.hs` file, so it uses as much of the `Tuura.Concept.STG.Translation` as possible. However, testing this produces a seg. fault, and debugging proves it is because `Signal` is defined in the `Translation` module, not in the `Main.hs` file. 

I have had a go at hiding `Signal` when importing `Tuura.Concept.STG.Translation`, and redefining it in `translate/Main.hs`, but this provides incompatability issues when attempting to retrieve information on the `CircuitConcept` itself. 

Thus, there is now a lot of repeated code, and changing the actual translation code used from `translate/Main.hs` means that the `TransTest.hs` file won't actually correctly test the translation process.

Do you have any ideas for this?


TransTest.hs
```diff
@@ -0,0 +1,131 @@
...
+    testOrGateConcept
+    testTwoAndGates
+    testTwoDifferentCElements
+
+testDotGOutput :: IO ()
+testDotGOutput = do
+    putStrLn "=== testDotGOutput"
+    assertEq threeSignalsConcept expected
+  where
+    expected = ".model out\n"          ++
```
snowleopard: Not that this is a bit difference, but I think `unlines $ [ ".model out", ...  ]` is typically easier to work with.

TransTest.hs
```diff
@@ -0,0 +1,131 @@
...
+               "C1 C-\n"               ++
+               "C- C0\n"               ++
+               ".marking {A0 B1 C0}\n" ++
+               ".end\n"
+
+testHandshakeConcept :: IO ()
+testHandshakeConcept = do
+    putStrLn "===testHandshakeConcept"
+    assertEq (sort handshakeConcept) (sort expected)
+  where
```
snowleopard, line 50: Let's put the definition of `handshakeConcept` right here. I see no reason to keep it in a separate function -- it just makes it harder to understand what is being tested.

Same for other concepts that we are testing.

TransTest.hs
```diff
@@ -0,0 +1,131 @@
...
+
+twoAndGatesConcept :: [([Transition Signal], Transition Signal)]
+twoAndGatesConcept = arcs circuit
+  where
+    circuit = inputs [a, b, c, d] <> outputs [out] <> initialise0 [a, b, c, d, out] <> andGate a b out <> andGate c d out
+    a   = Signal 0
+    b   = Signal 1
+    c   = Signal 2
+    d   = Signal 3
+    out = Signal 4
```
snowleopard: This could be written a little more compactly: `[a, b, c, d, out] = map Signal [0..4]`.

snowleopard (changes requested):
Regarding duplicated code: most of it seems to be generic enough, that is, it doesn't necessarily needs to use `Signal`s, instead it could be made abstract and use a type variable `a`. Then it could be moved to a separate file and imported in two places.

jrbeaumont:
@snowleopard I have made the changes to the test file. 

With regards to using a type variable, `[a]`. Most of it requires it to be `Signal`, as it needs to get information from the `Concept`. Are you sure we can use `[a]` instead?

TransTest.hs
```diff
@@ -18,61 +18,50 @@ testDotGOutput = do
...
+    assertEq (sort (arcs twoAndGatesConcept)) (sort expected)
   where
+    twoAndGatesConcept = inputs [a, b, c, d] <> outputs [out] <> initialise0 [a, b, c, d, out] <> andGate a b out <> andGate c d out
     expected = [([rise a], rise out), ([rise b], rise out), ([rise c], rise out), ([rise d], rise out), ([fall a, fall b], fall out), ([fall c, fall d], fall out)]
-    a   = Signal 0
-    b   = Signal 1
-    c   = Signal 2
-    d   = Signal 3
-    out = Signal 4
+    [a, b, c, d, out] = map Signal [0..4]
```
snowleopard, line 71: Perhaps, this could be a top-level declaration, then you don't need to declare signals in every test?

TransTest.hs
```diff
@@ -18,61 +18,50 @@ testDotGOutput = do
...
-               ".marking {A0 B1 C0}\n" ++
-               ".end\n"
+    threeSignalsConcept = doTranslate signs (inputs [a] <> outputs [b] <> internals [c] <> initialise0 [a, c] <> initialise1 [b])
+    a = Signal 0
+    b = Signal 1
+    c = Signal 2
+    signs = [a, b, c]
+    expected = unlines ([".model out", ".inputs A", ".outputs B", ".internal C", ".graph"] ++
+                        ["A0 A+", "A+ A1", "A1 A-", "A- A0", "B0 B+", "B+ B1", "B1 B-", "B- B0"] ++
+                        ["C0 C+", "C+ C1", "C1 C-", "C- C0", ".marking {A0 B1 C0}", ".end"])
```
snowleopard: Ah, I didn't mean to compress lines together -- sorry should have explained better :) Something like this:

``` haskell
    expected = unlines [ ".model out"
                       , ".inputs A"
                       , ".outputs B"
                       ... ]
```

snowleopard:
@jrbeaumont Let's look at function `doTranslate` for example. Can't it have the following type instead?

``` haskell
doTranslate :: (Show a, Ord a) => [a] -> Concept (State a) (Transition a) a -> String
```

As far as I can see, all we need is to `show` and `compare` signals?

jrbeaumont:
@snowleopard: OK I just found that as you posted it haha. Changed some things around, seems to work fine. Just going to test it now.

jrbeaumont:
@snowleopard: I have changed the code as needed, and tested it both alone, and in Workcraft, all appears to be working well.

src/Tuura/Concept/STG/Translation.hs
```diff
@@ -17,11 +17,7 @@ instance Ord Signal
     where
         compare (Signal x) (Signal y) = compare x y
 
-circuitName, tmpModuleFile :: String
-circuitName   = "circuit"
-tmpModuleFile = ".Helper.hs"
-
-doTranslate :: [Signal] -> Concept (State Signal) (Transition Signal) Signal -> String
+doTranslate :: (Show a, Ord a) => [a] -> Concept (State a) (Transition a) a -> String
```
snowleopard: I think `Concept (State a) (Transition a) a` is simply `CircuitConcept a`?

src/Tuura/Concept/STG/Translation.hs
```diff
@@ -51,7 +47,7 @@ validate signs circuit
     inconsistent = filter ((==Inconsistent) . initial circuit) signs
     undef        = filter ((==Undefined) . initial circuit) signs
 
-handleArcs :: [([Transition Signal], Transition Signal)] -> [String]
+handleArcs :: Show a => [([a], a)] -> [String]
```
snowleopard: I suggest to use `Transition a` here instead of just `a`. We don't have to make the code as abstract as possible, but abstract enough so we don't have to duplicate things. I think `Transition` does't get in the way of reusing code. In fact, it acts as documentation (clarifies the meaning of the code), plus allows you to use things like `newValue`.

src/Tuura/Concept/STG/Translation.hs
```diff
@@ -71,21 +67,21 @@ genSTG inputSigns outputSigns internalSigns arcStrs initStrs =
         allArcs = concatMap consistencyLoop allSigns ++ arcStrs
         marks = initVals allSigns initStrs
 
-addConsistencyTrans :: Transition Signal -> Int -> [String]
+addConsistencyTrans :: Show a => a -> Int -> [String]
 addConsistencyTrans effect n
-        | newValue effect = map (\x -> (printf "%s0 %s/%s\n" (init (show effect)) (show effect) (show x))
```
snowleopard: Having `Transition a` would help keep `newValue`.

src/Tuura/Concept/STG/Translation.hs
snowleopard: Also, I think `translate` would be a better name.

snowleopard (changes requested):
Thanks! Almost there :)

jrbeaumont:
Cleared those up, still working well :)
Let me know if you have more suggestions!

snowleopard (approved):

snowleopard:
@jrbeaumont Great, merged now!

jrbeaumont:
@snowleopard: Thanks!

PR #56, opened 2016-10-26
Changed stack resolver to work with Ubuntu

jrbeaumont:
Stack uses the `lts-6.20` resolver, and thus, `hint-5.2.0`.

Testing has proven no issues.


snowleopard:
Thanks!

PR #59, opened 2016-10-31
Stack should now use lts-7.5 for GHC 8.0.1

jrbeaumont:
Hopefully this should work correctly. 

I have no idea how that happened, sorry about that.


snowleopard:
Thanks @jrbeaumont!

PR #63, opened 2017-02-10
Added options for including external concept files

jrbeaumont:


translate/Main.hs
```diff
@@ -14,19 +14,21 @@ import Tuura.Concept.STG
...
 main = do
-    args <- getArgs
-    if length args /= 1
-        then putStrLn "Exactly one path needed"
-        else do
-            r <- GHC.runInterpreter $ doWork (head args)
-            case r of
-               Left err -> putStrLn $ displayException err
-               Right () -> return ()
+    -- args <- getArgs
```
snowleopard: Do we really need this comment?

jrbeaumont: Nope, sorry, left over from refactoring.

translate/Main.hs
```diff
@@ -14,19 +14,21 @@ import Tuura.Concept.STG
...
+    options <- getOptions
+    let input = optInput options
+    let paths = [input] ++ optInclude options
+    r <- GHC.runInterpreter $ doWork paths
+    case r of
+        Left err -> putStrLn $ displayException err
+        Right () -> return ()
```
snowleopard: You can simplify the above three lines as `either (putStrLn . displayException) return r`  although I admit that may be more cryptic :) If you'd like to stick to your code, let me know and I'll merge.

snowleopard:
Nice. Thanks @jrbeaumont!

You can probably merge this yourself now? If yes, feel free to go ahead :)

translate/Main.hs
jrbeaumont: I will give this a try, compact code is always nice :)

snowleopard: `either` is a useful function when working with `Either`. It does either this or that :)

jrbeaumont:
I can merge myself but I'll make some small changes and sort it out :)

Adrian has completed the FSM translation so I'm gonna merge it and test and do another PR for that, which would be nice to have some of your input for that. 

I'm thinking we should have separate FSM and STG libraries, as at some point they may diverge, with things like the unstable signal implementation, or anything else which may not be transferable for FSMs and STGs and it may be better to have separate concept libraries for these from the beginning, instead of combining these, and then separating them again later. What do you think?

snowleopard:
OK, makes sense. Let's keep FSM and STG libraries separate for now. But can you do that without duplicating some big chunks of code, like the Cartesian product stuff?

Happy to do code review!

jrbeaumont:
I am not sure how much I can avoid duplicating. I will have a play around and see what's is possible.

PR #66, opened 2017-02-20
Refactored repeated circuit/abstract code into single files

jrbeaumont:
@snowleopard: I have fixed the repeated code issues.

I had to create `Tuura.Concept.Circuit...` as I tried replacing the existing `Tuura.Concept.Circuit` and `Tuura.Concept.Abstract` but this causes issues with the simulation testing.

What do you think now?


snowleopard:
Looks good! 

To avoid awkward `Circuit.Circuit`, perhaps, we could have `Circuit.Basic` (instead of `Circuit.Abstract`) and `Circuit.Derived` (instead of `Circuit.Circuit`)?

jrbeaumont:
@snowleopard I have renamed those. They are better names. Thank you

snowleopard:
Thanks! I have a feeling that we will still need to make some renaming to finally get it right, but for now let's go with this.

Shall I merge?

jrbeaumont:
Not yet, there's issues with the testing that I have just fixed. Let's just confirm that the CI passes :)

I agree, some names need to be changed, but we can work this out while the tool works.

jrbeaumont:
You can merge this now if you'd like :)

jrbeaumont:
@snowleopard Can we merge this now?

snowleopard:
@jrbeaumont Done!

PR #67, opened 2017-02-23
Added support for invariants

jrbeaumont:
This includes:

* Support for invariants as a part of `CircuitConcept`.

* Translating FSMs will no longer insert invariant states.

* Translating STGs will report on what sets of transitions must occur to reach invariant states.

* No translation will occur if the provided initial state is an invariant state.


.gitignore
```diff
@@ -3,4 +3,6 @@
 *.o
 dist/
 .stack-work/
-*.DS_Store
\ No newline at end of file
+<<<<<<< HEAD
+*.DS_Store
```
snowleopard: What's going on here?

TransTest.hs
```diff
@@ -43,7 +43,8 @@ testDotGOutput = do
                        , "C1 C-"
                        , "C- C0"
                        , ".marking {A0 B1 C0}"
-                       , ".end"]
+                       , ".end"
+                       , ""]
```
snowleopard: Why do we need an extra empty line in the end?

.gitignore
jrbeaumont: Sorry that's my mistake. `*.DS_store` occurs in all directories on macs, and git doesn't auto-ignore it. I didn't mean to commit it, it's just an annoying thing that always shows up in git!

TransTest.hs
jrbeaumont: This matches there being no output for mentioning invariant states.

src/Tuura/Concept/Circuit/Basic.hs
```diff
@@ -35,17 +36,19 @@ data Concept s e a = Concept
                    {
                        initial   :: a -> InitialValue,
                        arcs      :: [([e], e)],
-                       interface :: a -> Interface
+                       interface :: a -> Interface,
+                       invariant :: [[e]]
                    }
```
snowleopard: So, this is a list of lists of events. What is its meaning? Maybe it's worth adding a proper datatype for better clarity. For example, `invariant :: [Invariant e]`, that is, we have a list of `Invariant` descriptions. 

In fact, we could do something similar with `arcs` at some point. For example, `arcs :: [Causality e]`, and then define `data Causality e = AndCausality e e | OrCausality [e] e`. That would likely make the code cleaner and self-explanatory.

src/Tuura/Concept/Circuit/Derived.hs
```diff
@@ -106,7 +106,10 @@ handshake11 a b = handshake a b <> initialise a True <> initialise b True
 
 -- TODO: Restrict the initial state so that a=b=1 is not allowed.
 me :: a -> a -> CircuitConcept a
-me a b = fall a ~> rise b <> fall b ~> rise a
+me a b = fall a ~> rise b <> fall b ~> rise a <> invariantConcept [rise a, rise b]
```
snowleopard: This is very confusing: the most likely interpretation of `invariantConcept [rise a, rise b]` is that we always have `a = b = 1`. Clearly, the meaning is the opposite. So, we better use something like `never [rise a, rise b]` here.

src/Tuura/Concept/FSM/Translation.hs
```diff
@@ -205,6 +210,11 @@ fsmarcxToFsmarc arc = FsmArc newSourceEnc (transx arc) newDestEnc
     where newSourceEnc = (encToInt . srcEncx) arc
           newDestEnc = (encToInt . destEncx) arc
 
+removeInvariants :: [FsmArcX a] -> [[Tristate]] -> [FsmArcX a]
+removeInvariants xs inv = concatMap (\x -> if (destEncx x `elem` inv || srcEncx x `elem` inv)
+                          then []
+                          else [x]) xs
+
```
snowleopard: It looks like you should have used `filter` here: `filter (\x -> ...) xs`?

src/Tuura/Concept/FSM/Translation.hs
```diff
@@ -205,6 +210,11 @@ fsmarcxToFsmarc arc = FsmArc newSourceEnc (transx arc) newDestEnc
...
 -- Produce all arcs with all X's resolved
-createAllArcs :: Ord a => [([Transition a], Transition a)] -> [FsmArc a]
-createAllArcs = map fsmarcxToFsmarc . expandAllXs . createArcs
+createAllArcs :: Ord a => [([Transition a], Transition a)] -> [[Tristate]] -> [FsmArc a]
+createAllArcs xs inv = map (fsmarcxToFsmarc) (removeInvariants (expandAllXs (createArcs xs)) inv)
```
snowleopard: Can this be simplified as `createAllArcs xs = map fsmarcxToFsmarc . removeInvariants (expandAllXs (createArcs xs))`?

src/Tuura/Concept/STG/Translation.hs
```diff
@@ -94,3 +85,10 @@ initVal s ls = sum (map (\x -> if (fst x == s) then fromEnum (snd x) else 0) ls)
 
 readArc :: String -> String -> [String]
 readArc f t = [f ++ " " ++ t, t ++ " " ++ f]
+
+genInvStrs :: (Eq a, Show a) => [[Transition a]] -> [String]
+genInvStrs invs
+        | invs      == [] = []
+        | otherwise = info ++ map (\i -> unwords (map show i)) invs
+    where
+        info = ["Signal transition sets which are invariant:"]
```
snowleopard: To preserve compatibility with existing tools, I suggest you write invariant info in comments, and before the `.end`. For example: `# invariant = not (X && Y)`.

src/Tuura/Plato/Translation.hs
```diff
@@ -1,8 +1,12 @@
 module Tuura.Plato.Translation where
 
 import Data.Char
+import Data.List.Extra
 
-data ValidationResult a = Valid | Invalid [a] [a] [a] deriving Eq
+import Tuura.Concept.Circuit.Basic
+import Tuura.Concept.Circuit.Derived
+
+data ValidationResult a = Valid | Invalid [a] [a] [a] [a] deriving Eq
```
snowleopard: This is getting too hard to read. Try to improve this datatype.

src/Tuura/Plato/Translation.hs
```diff
@@ -15,12 +19,36 @@ instance Ord Signal
...
+    undef        = filter ((==Undefined) . initial circuit) signs
+    invInit      = checkInitialStates signs circuit
+
+checkInitialStates :: Ord a => [a] -> CircuitConcept a -> [a]
+checkInitialStates signs circuit = nubOrd (concatMap (\inv -> if (checkForInvariant inv initialStates) then (invariantError inv) else []) (invariant circuit))
+  where
+    initialStates = map (\s -> Transition { signal = s , newValue = getDefined $ initial circuit s }) signs
+
+invariantError :: [Transition a] -> [a]
+invariantError = map (signal)
```
snowleopard: No need for `()` here.

src/Tuura/Plato/Translation.hs
```diff
@@ -15,12 +19,36 @@ instance Ord Signal
...
+
+checkForInvariant :: Eq a => [Transition a] -> [Transition a] -> Bool
+checkForInvariant invs initialStates
```
snowleopard: Rewrite this without explicit recursion. Check out standard `all` and `any` functions in Haskell.

snowleopard:
@jrbeaumont I've added a bunch of comments. Let me know when you are ready for another iteration.

.gitignore
snowleopard: I'm fine to have `*.DS_Store` in `.gitignore`. The problem is the `<<<<<<< HEAD` thing.

TransTest.hs
snowleopard: See my other comment. I think we should write invariants in comments, before the `.end` line.

src/Tuura/Concept/Circuit/Basic.hs
jrbeaumont: That would work better. I will start by doing it for Invariant, then for Causality later.

src/Tuura/Concept/Circuit/Derived.hs
jrbeaumont: I may keep invariant concept in `Basic` as this matches the names of the other standard concept types. I will implement a `never` function though for this purpose.

src/Tuura/Concept/FSM/Translation.hs
jrbeaumont: That would be better, yes. Sorry, I'm still getting the hang of useful Haskell functions.

src/Tuura/Concept/FSM/Translation.hs
jrbeaumont: More than likely. I will give it a try, and find it's most simplified form, for less confusion.

src/Tuura/Concept/STG/Translation.hs
jrbeaumont: Would this not get lost in the parsing of the `.g` file though? For example, in Workcraft, do we not want this information to be output to the user?

src/Tuura/Plato/Translation.hs
jrbeaumont: I'm not overly sure how we can do that. It makes sense, in that it's invalid. 
I suppose I could create a data type, say `Errors a` which holds these 4 pieces of information?
Or maybe split off inconsistent and undefined initial states into some form of `InitialError` type, which is handled within itself?

src/Tuura/Concept/Circuit/Basic.hs
snowleopard: I think this could work very nicely as a definition for `Invariant`: 
```haskell
data Invariant e = NeverAll [e]

never es = invariantConcept [NeverAll es] 
```

That is, at the moment, the only type of invariants we support are `NeverAll` invariants. We could extend this list later.

src/Tuura/Concept/STG/Translation.hs
snowleopard: Yes, the tools that are not ready for this, will simply ignore such lines. But the point is that we can support this by modifying the parser of `.g` files in Workcraft so that it parses these lines and stores them appropriately within the model.

src/Tuura/Plato/Translation.hs
snowleopard: What about:
```haskell
data ValidationResult a = Valid | Invalid [ValidationError a] deriving Eq

data ValidationError a = UnusedSignal a | InconsistentInitialState a | ...
```

jrbeaumont: Yeah that's what I was thinking. I'll sort this out.

.gitignore
jrbeaumont: Yeah it was from an error I tried to not include haha. I'll fix it.

src/Tuura/Concept/Circuit/Basic.hs
jrbeaumont: That's fair enough. I was planning on doing something similar. I will include this in the next iteration.

src/Tuura/Concept/STG/Translation.hs
jrbeaumont: Oh right I see. I can arrange this then.

jrbeaumont:
@snowleopard: Here is iteration 2 :)

src/Tuura/Concept/Circuit/Derived.hs
```diff
@@ -106,10 +106,10 @@ handshake11 a b = handshake a b <> initialise a True <> initialise b True
 
 -- TODO: Restrict the initial state so that a=b=1 is not allowed.
```
snowleopard: The TODO can now be dropped.

src/Tuura/Concept/STG/Translation.hs
```diff
@@ -69,7 +69,7 @@ transition (f, t)
         | otherwise  = readArc (init (show f) ++ "0") t
 
 tmpl :: String
-tmpl = unlines [".model out", ".inputs %s", ".outputs %s", ".internal %s", ".graph", "%s.marking {%s}", ".end", "%s"]
+tmpl = unlines [".model out", ".inputs %s", ".outputs %s", ".internal %s", ".graph", "%s.marking {%s}", "%s.end"]
```
snowleopard: Shouldn't the last element be turned into two `"%s", ".end"`?

jrbeaumont: Having two elements add's an unnecessary new line. 

I can remove this by adding a check somewhere for the last element in a list, but that seems like a lot more code.

src/Tuura/Concept/STG/Translation.hs
```diff
@@ -86,9 +86,10 @@ initVal s ls = sum (map (\x -> if (fst x == s) then fromEnum (snd x) else 0) ls)
 readArc :: String -> String -> [String]
 readArc f t = [f ++ " " ++ t, t ++ " " ++ f]
 
-genInvStrs :: (Eq a, Show a) => [[Transition a]] -> [String]
-genInvStrs invs
-        | invs      == [] = []
-        | otherwise = info ++ map (\i -> unwords (map show i)) invs
+genInvStrs :: (Ord a, Show a) => Invariant (Transition a) -> String
+genInvStrs (NeverAll es)
```
snowleopard: Again, get rid of explicit recursion -- it's almost never required in Haskell because there are so many powerful combinators.

In this case you most probably need to first `map format es` and then do something like `intercalate "&&"`.

src/Tuura/Concept/STG/Translation.hs
```diff
@@ -86,9 +86,10 @@ initVal s ls = sum (map (\x -> if (fst x == s) then fromEnum (snd x) else 0) ls)
...
+        | es        == [] = []
+        | otherwise = "# invariant = not (" ++ (format (head sorted)) ++ (concatMap (\e -> " && " ++ format e) (tail sorted)) ++ ")"
     where
-        info = ["Signal transition sets which are invariant:"]
+        format e = if (newValue e) then show (signal e) else "not " ++ show (signal e)
+        sorted = nubOrd es
```
snowleopard: Are you using `nubOrd` to simplify the invariant? This looks correct, but I'm not sure we should do that -- let's make this function simpler -- you get what you started with (may be helpful for debugging, for example).

src/Tuura/Plato/Translation.hs
```diff
@@ -6,7 +6,13 @@ import Data.List.Extra
 import Tuura.Concept.Circuit.Basic
 import Tuura.Concept.Circuit.Derived
 
-data ValidationResult a = Valid | Invalid [a] [a] [a] [a] deriving Eq
+data ValidationResult a = Valid | Invalid [ValidationError a] deriving Eq
+
+data ValidationError a = UnusedSignal [a]
```
snowleopard: I think `UnusedSignal a` is more precise. If there are two undefined signals, you simply do `Invalid [UnusedSignal a, UnusedSignal b]`. With you current approach there are multiple ways to achieve the same thing, e.g. `Invalid [UnusedSignal [a, b]]`.

src/Tuura/Concept/STG/Translation.hs
jrbeaumont: Having just looked it up, `intersperse` would work better!

src/Tuura/Plato/Translation.hs
```diff
@@ -19,36 +25,50 @@ instance Ord Signal
...
+  --   ic = if (unused  /= []) then ("The following signals have inconsistent inital states: \n" ++ unlines (map show incons) ++ "\n") else ""
+  --   ud = if (undef   /= []) then ("The following signals have undefined initial states: \n" ++ unlines (map show undef) ++ "\n") else ""
+  --   iv = if (invInit /= []) then ("The initial state is an invariant state.\nSignals with erroneous initial states: \n" ++ unlines (map show invInit) ++ "\n") else ""
 
 validate :: Ord a => [a] -> CircuitConcept a -> ValidationResult a
 validate signs circuit
     | unused ++ inconsistent ++ undef ++ invInit == [] = Valid
-    | otherwise                             = Invalid unused inconsistent undef invInit
+    | otherwise = Invalid (
+        (if unused       /= [] then [UnusedSignal unused]                   else []) ++
```
snowleopard: With the above comment, you would be able to write `map UnusedSignal unused` instead of this line.

src/Tuura/Concept/STG/Translation.hs
jrbeaumont: It's used to remove duplicates. It's not entirely necessary I suppose. I can remove it.

src/Tuura/Plato/Translation.hs
```diff
@@ -19,36 +25,50 @@ instance Ord Signal
...
 
 invariantError :: [Transition a] -> [a]
-invariantError = map (signal)
+invariantError = map signal
 
 checkForInvariant :: Eq a => [Transition a] -> [Transition a] -> Bool
-checkForInvariant invs initialStates
-    | invs == [] = True
-    | otherwise  = ((head invs) `elem` initialStates) && (checkForInvariant (tail invs) initialStates)
+checkForInvariant invs initialStates = all (`elem` initialStates) invs
```
snowleopard: Much nicer! :-)

src/Tuura/Plato/Translation.hs
jrbeaumont: That's fair enough, I can see how it's not as precise.

src/Tuura/Concept/STG/Translation.hs
snowleopard: I see. Then you can probably leave this as is.

src/Tuura/Plato/Translation.hs
jrbeaumont: I agree, what a nice simple function :) I might remove the function actually, and just have:
`all elem initialStates` instead.

snowleopard:
Going in the right direction... :)

src/Tuura/Plato/Translation.hs
snowleopard: Indeed, it makes sense. Do remove it!

src/Tuura/Concept/STG/Translation.hs
jrbeaumont: I was wrong, `intersperse` can't take in `"&&"` as this is a list.
`intercalate` was correct!

snowleopard: Ah, glad that my hunch was correct ;)

jrbeaumont: I got cocky haha.

jrbeaumont:
@snowleopard: Are you aware of a way to filter by pattern? I have changed `UnusedSignal [a]` etc to `UnusedSignal a`, but I need to map all the signals by their patterns, only if there are those types of errors. 
Doing `filter (==UnusedSignal) errors` or anything similer, such as `==(UnusedSignal a)` will not work.

snowleopard:
@jrbeaumont I don't understand, can you give an example?

jrbeaumont:
So I have a list: `[ValidationError a]`
In this case, let's say it consists of: `[UnusedSignal x, UndefinedInitialState x, InvariantInitialState x, InvariantInitialState y]`
`x` and `y` are the erroneous signals.

When it comes to preparing the outputs to inform the users what signals are erroneous, I would like to filter by all the `ValidationError a` types, so I can have:

`[UnusedSignal x]`
`[UndefinedInitialState x]`
`[InvariantInitialState x, InvariantInitialState y]`

This will then allow me to add the information on what is wrong with the signals, i.e. "The following signals are not defined is inputs, outputs or internals", and then list the signals, in this case `x`.
Then I can do the same for `UndefinedInitialState`, in this case `x` again, `InconsistentInitialState` will not have any signals so no output will be produced for this, and for `InvariantInitialState` the output will include `x` and `y`.

My issue is, because these are types including arguments (Each type includes `a`, i.e. `UnusedSignal a`) I cannot filter the full list of `[ValidationError a]` to simply provide me with lists for each of these types. How can I filter for each type of `ValidationError a`

snowleopard:
@jrbeaumont I see, thanks! I think the following list comprehension trick should work: 
```haskell
unused :: [ValidationError a] -> [a]
unused es = [ a | UnusedSignal a <- es ]
```
In list comprehensions, if the pattern match fails the current element is skipped.

See https://en.wikibooks.org/wiki/Haskell/Pattern_matching#List_comprehensions.

jrbeaumont:
That's great, thank you!

jrbeaumont:
@snowleopard: OK, here is the third iteration :)

jrbeaumont:
@snowleopard: As the latest commits suggest, only reachable states will be inserted for FSMs.

If any non-invariant states are reachable an error will be produced instead of an FSM.
A warning will be given if any states are not reachable, but are not declared as non-invariant.
If `reachability=inv`, then this information is printed. 

Please have a look when you get time and let me know!

Edit: Am I explaining this all correctly? I've spent so long staring at the words invariant and non-invariant I've forgotten which is which :|

src/Tuura/Concept/FSM/Translation.hs
```diff
@@ -82,20 +82,28 @@ translate :: (Show a, Ord a) => CircuitConcept a -> [a] -> String
...
-          let sortedArcs = concatMap handleArcs (groupSortOn snd allArcs)
-          let invariants = concatMap (\i -> createInvariants i signs) (invariant circuit)
-          let arcStrs = map show (createAllArcs sortedArcs invariants)
+          let allCause = addConsistency (arcs circuit) signs
+          let sortedCause = concatMap handleArcs (groupSortOn snd allCause)
+          let initialState = getInitialState circuit signs
+          let allArcs = createAllArcs sortedCause
+          let reachables = findReachables allArcs initialState
+          let invariants = map encToInt (concatMap (\i -> getInvariantStates i signs) (invariant circuit))
+          let reachableArcs = removeUnreachables allArcs reachables
```
snowleopard: Let's drop all the `let`s but the first one.

snowleopard: (Throughout the code)

src/Tuura/Concept/FSM/Translation.hs
```diff
@@ -82,20 +82,28 @@ translate :: (Show a, Ord a) => CircuitConcept a -> [a] -> String
...
+              genFSM inputSigns outputSigns internalSigns (map show reachableArcs) (show initialState) reachReport
       Invalid errs ->
           "Error. \n" ++ addErrors errs
 
-getInitialState :: CircuitConcept a -> [a] -> String
-getInitialState circuit signs = show (encToInt state)
+getInitialState :: CircuitConcept a -> [a] -> Int
+getInitialState circuit signs = encToInt state
   where
     state = map (\s -> if (getDefined $ initial circuit s) then triTrue else triFalse) signs
```
snowleopard: I think it may be useful to define a function `fromBool :: Bool -> Tristate` and use it here.

src/Tuura/Concept/FSM/Translation.hs
jrbeaumont: I tried this and it didn't compile :S

src/Tuura/Concept/FSM/Translation.hs
```diff
@@ -110,16 +118,31 @@ handleArcs arcLists = result
...
+      outs = map show outputSigns
+      ins = map show inputSigns
+      ints = map show internalSigns
+
+reachableInvariantStateError :: Show a => [a] -> String
+reachableInvariantStateError es = "Error:\n" ++
+                                  "The following state(s) are reachable but the invariant does not hold for them:\n" ++
+                                  unlines (errStates)
+    where
+      errStates = map (\e -> "s" ++ (show e)) es
```
snowleopard: In such cases list comprehension usually works better: `errStates = [ "s" ++ show e | e <- es ]`.

src/Tuura/Concept/FSM/Translation.hs
jrbeaumont: Nevermind, I see what I did wrong.

src/Tuura/Concept/FSM/Translation.hs
```diff
@@ -209,9 +232,22 @@ fsmarcxToFsmarc arc = FsmArc newSourceEnc (transx arc) newDestEnc
...
+createAllArcs :: Ord a => [([Transition a], Transition a)] -> [FsmArc a]
+createAllArcs = map fsmarcxToFsmarc . expandAllXs . createArcs
+
+findReachables :: [FsmArc a] -> Int -> [Int]
+findReachables allArcs initialState = nubOrd (visit initialState allArcs [])
+
+visit :: Int -> [FsmArc a] -> [Int] -> [Int]
+visit state allArcs visited = [state] ++
+        if (nonVisited /= [])
+        then concatMap (\s -> visit s allArcs (visited ++ [state] ++ nonVisited)) nonVisited
```
snowleopard: Are you sure you need `if-then-else` here? Looks like you can delete it :)

src/Tuura/Concept/FSM/Translation.hs
```diff
@@ -209,9 +232,22 @@ fsmarcxToFsmarc arc = FsmArc newSourceEnc (transx arc) newDestEnc
...
+        else []
+    where
+      indices = elemIndices state (map srcEnc allArcs)
+      destArcs = map ((map destEnc allArcs) !!) indices
+      nonVisited = filter (\x -> (x `notElem` visited)) destArcs
```
snowleopard: Should this be ``filter (`notElem` visited) destArcs``?

src/Tuura/Plato/Translation.hs
```diff
@@ -40,7 +40,7 @@ addErrors errs =
              ++ unlines (map show (undefd errs)) ++ "\n"
         else "") ++
         (if (invart errs) /= []
-        then "The initial state is an invariant state.\nSignals with erroneous initial states: \n"
+        then "The invariant does not hold for the initial state.\nSignals with erroneous initial states: \n"
              ++ unlines (map show (invart errs)) ++ "\n"
         else "")
```
snowleopard: I'm sure you can somehow get rid of `else ""` here.

src/Tuura/Concept/FSM/Translation.hs
```diff
@@ -209,9 +232,22 @@ fsmarcxToFsmarc arc = FsmArc newSourceEnc (transx arc) newDestEnc
...
+findReachables allArcs initialState = nubOrd (visit initialState allArcs [])
+
+visit :: Int -> [FsmArc a] -> [Int] -> [Int]
+visit state allArcs visited = [state] ++
+        if (nonVisited /= [])
+        then concatMap (\s -> visit s allArcs (visited ++ [state] ++ nonVisited)) nonVisited
+        else []
+    where
+      indices = elemIndices state (map srcEnc allArcs)
+      destArcs = map ((map destEnc allArcs) !!) indices
```
snowleopard: `!!` is very slow. Perhaps, you could implement this using a `Data.Set` or `Data.Map`?

src/Tuura/Concept/FSM/Translation.hs
```diff
@@ -165,8 +188,8 @@ createArcs xs = zipWith3 createArc makeSrcEncs makeDestEncs activeTransitions
           invert = liftM2 Transition signal (not . newValue)
           activeTransitions = (map snd .  addMissingSignals) xs
 
-createInvariants :: Ord a => Invariant (Transition a) -> [a] -> [[Tristate]]
-createInvariants (NeverAll es) allSigns = expand (encode newTransitions)
+getInvariantStates :: Ord a => Invariant (Transition a) -> [a] -> [[Tristate]]
+getInvariantStates (NeverAll es) allSigns = expand (encode newTransitions)
     where newTransitions = transX ++ (concatMap (map (\s -> TransitionX { msignal = s, mnewValue = triX })) missingSignals)
           transX = map toTransitionX es
           missingSignals = map (allSigns \\) (onlySignals [es])
```
snowleopard: This look suspicious :)

src/Tuura/Plato/Translation.hs
```diff
@@ -25,50 +25,48 @@ instance Ord Signal
...
-        UnusedSignal as ->
-            "The following signals are not declared as input, output or internal: \n" ++ unlines (signs as) ++ "\n"
-        InconsistentInitialState as ->
-            "The following signals have inconsistent inital states: \n" ++ unlines (signs as) ++ "\n"
-        UndefinedInitialState as ->
-            "The following signals have undefined initial states: \n" ++ unlines (signs as) ++ "\n"
-        InvariantInitialState as ->
-            "The initial state is an invariant state.\nSignals with erroneous initial states: \n" ++ unlines (signs as) ++ "\n"
+addErrors :: (Eq a, Show a) => [ValidationError a] -> String
+addErrors errs =
```
snowleopard: The code below looks ugly. Try to improve.

snowleopard:
@jrbeaumont I left some comments. Some of the comments are very non-constructive, sorry, but I didn't have time to investigate further :-)

> Edit: Am I explaining this all correctly? I've spent so long staring at the words invariant and non-invariant I've forgotten which is which :|

Yes, I think in this version "invariant" is used correctly :)

Although I don't think the word is used as an adjective, so "non-invariant state" feels strange. I'd say "state where the invariant doesn't hold" -- a lot more verbose, but sounds familiar.

jrbeaumont:
Thank you @snowleopard, I will get to these comments :)

I see what you mean about it not being an adjective, but I'm considering renaming the error type `InvariantInitialState` to `NonInvariantInitialState` so it makes more sense, as if the initial state holds the invariant, an error calling it "InvariantInitialState" may be confusing. What do you think?

snowleopard:
Ah, I missed that. `InvariantInitialState` is indeed incorrect. But now, since you are reporting an error if the invariant is violated in general (not only in the initial state), why not call it `InvariantViolated`?

jrbeaumont:
That's a good idea. However, for the invariants being reachable in FSMs, this is only recognised after the validation. I can create this as this type of error anyway, for clarity, but it will have to be dealt with in the same way as it is in this iteration.

snowleopard:
> However, for the invariants being reachable in FSMs, this is only recognised after the validation

Why? If the invariant is broken, the specification is invalid. So, I don't see why this should be treated differently from other errors.

jrbeaumont:
This happens because the validation occurs, then if this is valid, the FSM translation starts, where we then find the reachable states, which then throws an error, but we cannot return an `invalid` from the translate function.

snowleopard:
Well, still it doesn't make sense to verify the initial state separately, if all states are being verified. Just remove this check altogether.

jrbeaumont:
That's a good idea. I will have a think and see if there is a way to check for reachable states which don't hold for the invariant before we start any translation, so an `Invalid` is passed into the `translate` function of either `FSM` or `STG` translation files.

src/Tuura/Plato/Translation.hs
jrbeaumont: I've spent a while on this, and I cannot figure out how.

src/Tuura/Concept/FSM/Translation.hs
jrbeaumont: What do you recommend to improve it?

snowleopard: Nothing specific -- just looks like this could be optimised. If you can't -- leave it as is.

src/Tuura/Plato/Translation.hs
snowleopard: OK, then leave it as is for now.

src/Tuura/Concept/FSM/Translation.hs
jrbeaumont: It likely can. I'll have a try and if I can't I'll leave a TODO.

jrbeaumont:
@snowleopard: I have an idea for checking for reachable states which don't hold for the invariant in the validation function, before we try to translate anything. 
Please let me know what you think:

The idea is, we have all the arcs, consisting of `([causes], effect)`. 
If we have a `never` concept: `never [rise a, rise b]`, we can check the arcs.

Let's take the first transition of this `never`: `rise a`. 
Using this, let's filter all of the arcs where the `effect` is `rise a`.
Now, if we look at all of the causes for `rise a`, we can check to see if `fall b` is one of them, specifically, the opposite of the other transition of the `never` concept. If we find it, then this means that `b` must be 0 before `rise a` can occur.

If this is not found, we then check for `fall b` in all of the `causes` for these `causes` (`causes` where each `cause` of `rise a` is an `effect` in an arc). If we find `fall b` then no invariant is found.
Otherwise, use repeat this step with all the `causes` of these, and so on and so forth.

We can keep a list of visited `effect` transitions, so we don't repeat. If we never find this `fall b` transition, then then it is possible to reach a state where `rise a` and `rise b` have occurred. We can stop early if we find `rise b`.

I have a feeling I'm missing something with this, but even for lists of `causes` for OR causality, if any of the possible causes are `rise b`, it can reach a state that it shouldn't. 

We could even sequence the causes before hand, which would provide us with all of the necessary combinations of `causes` so we know exactly what we're dealing with. In the event there is some OR-causality which only one of the possible `causes` features `fall b` this will fail too. 

I might give it a try and see if I hit any roadblocks.

jrbeaumont:
@snowleopard: Thank you for all of these comments. If you have any further, let me know and I will get to them :)

Have you had chance to read my above comment?

snowleopard:
@jrbeaumont Ah, I missed the above comment. I suggest not to go that way for now. Let's fix minor things and commit this solution, perhaps, creating an issue if you'd like to improve something in a substantial way.

jrbeaumont:
No problem :)

That's fair. I would like to give it a try, as if we can check that no invariant states in `Plato.Translation` all errors will be found/reported before we even start to try and translate to an FSM/STG. I will fix any more changes you suggest, and I will make an issue.

snowleopard:
Merged now!

Issue #69, opened 2017-03-02
Refactor arc representation so AND and OR causality are more obvious.


jrbeaumont:
Currently, we represent all arcs as `arcs = [([e],e)]`. 

AND causality is included in this list by using a single item list for the cause transitions, i.e. 
```haskell
rise a ~> rise b = ([a+], b+)
```
OR causality is included using lists of transitions, i.e.
```haskell
[rise a,  rise b] ~> rise c = ([a+, b+], c+)
```

This is not the clearest of representations, so we should refactor.
A suggestion is:
```haskell
arcs :: [Causality e]
data Causality e = AndCausality e e | OrCausality [e] e. 
```
Here, `AndCausality` is simply the cause and effect transitions, `OrCausality` is the list of possible causes for the effect. We can still use this with the Cartesian Product to produce a correct STG/FSM.

snowleopard:
Another potential benefit of this approach is that is allows to introduce other constructors:
```haskell
data OrderConstraint a = AndCausality e e | OrCausality [e] e | TimingAssumption e e ...
```
In general, `OrderConstraint` may be a better name, as it captures not only causality, but other things.

jrbeaumont:
This is very true, if based on the type of a signal. It kills multiple birds with one stone

jrbeaumont:
A question though. What about OR causality with timing assumption, for example, input `a`  or input `b` transitioning high can cause input `c` to transition high. 

Is this feasible?

snowleopard:
You could probably come up with such an example, but I don't think I came across something like this in practice.

jrbeaumont:
That's what I was thinking. I don't think it's such a thing practically, it may be theoretically. In the event that such a timing assumption is found, we can add it in, but for now OR causality will cover it.

jrbeaumont:
We have changed the causality representation again (#75). This issue may therefore be closable. 
However, it is still possible to refactor this representation into something that may include assumptions etc, but we can cross this bridge at a later date. 

In summary, I suggest we close this issue.

snowleopard:
Agree, let's close this for now.

Issue #68, opened 2017-03-02
Check for reachable states which don't hold for the invariant before starting any translation


jrbeaumont:
At the moment, we check for reachable states in `FSM.Translation`, because we need to start the translation process to generate the states which we then search. We then use this information to see if any of the reachable states do not hold for the invariant. 

I suggest we provide some way of validating the reachable states before we enter `FSM.Translation`, or `STG.Translation` for that matter, preferably in `Plato.Translation`, as part of the `validate` function, as this will unify all the validation and error reporting.

jrbeaumont:
I have attempted to implement this (in the `reach` branch of my repo). Preliminary testing proves it to work, but I aim to do some more tests and then display the results here, for discussion before sending any PRs.

jrbeaumont:
OK, here is some testing! 

Test 1: Standard ME vs Never.
```haskell
circuit a b c = me a b 
                  <> initialise0 [a, b, c] <> inputs [a, b] <> outputs [c]
```
Mutual exclusion concepts includes `fall a ~> rise b <> fall b ~> rise a`, and thus, `never [rise a, rise b]` holds, as b cannot rise without a falling and vice versa. The output of this is: 
```
.model out
.inputs A B
.outputs C
.internal 
.graph
A0 A+
A+ A1
A1 A-
A- A0
B0 B+
B+ B1
B1 B-
B- B0
C0 C+
C+ C1
C1 C-
C- C0
B0 A+
A+ B0
A0 B+
B+ A0
.marking {A0 B0 C0}
# invariant = not (A && B)
.end
```
NOTE: This is just STG output, FSM output is also successful. This check occurs before any translation!

If we change this to
```haskell
circuit a b c = never [rise a, rise b] 
                  <> initialise0 [a, b, c] <> inputs [a, b] <> outputs [c]
```
We now don't include the `fall a ~> rise b <> fall b ~> rise a`, meaning that it is possible to reach `rise a, rise b`. 

Attempting to translate this gives us: 
```
Error. 
These signals violate the invariant: 
A
B
```

I think I am correct in assuming that this is correct. More testing incoming.

jrbeaumont:
Test 2: Adding a third signal to the never.

In this test, we have: 

```haskell
circuit a b c = never [rise a, rise b, rise c] 
                  <> initialise0 [a, b, c] <> inputs [a, b] <> outputs [c]
```
Which is basically the same as before, and thus should fail, so let's view the output:

```
Error. 
[...]

PR #71, opened 2017-03-07
Reordered validation in FSM translation and Refactored causality.

jrbeaumont:
So far, added `AndCausality` and `OrCausality` to `Causality` data type, which starts off #69, but we will need more discussion to include things like `TimingAssumption`.  

Validation and translation are reordered for FSMs (#68). The initial states are checked first, and if valid, the translation is done. Following this, a check for signals with no interface declared, as well as for reachable `never` states. If this is valid, then the translation completes, outputting the `.sg` file.  Otherwise, it outputs a handy error message, explaining either the problematic signal, or the `never` that is violated.

### DIFF
--- a/src/Tuura/Concept/Circuit/Basic.hs
+++ b/src/Tuura/Concept/Circuit/Basic.hs
@@ -15,7 +15,7 @@ import Data.Monoid
 -- * s is the type of states
 -- * e is the type of events
 
-data Causality e = AndCausality (e, e) | OrCausality ([e], e) deriving (Ord, Eq, Show)
+data Causality e = AndCausality e e | OrCausality [e] e deriving (Ord, Eq, Show)
 
 data Interface = Unused | Input | Output | Internal deriving (Ord, Eq, Show)
 
@@ -58,10 +58,10 @@ instance Monoid (Concept s e a) where
                   }
 
 arcConcept :: e -> e -> Concept s e a
-arcConcept from to = mempty { arcs = [AndCausality (from, to)] }
+arcConcept from to = mempty { arcs = [AndCausality from to] }
 
 orCausality :: [e] -> e -> Concept s e a
-orCausality from to = mempty { arcs = [OrCausality (from, to)] }
+orCausality from to = mempty { arcs = [OrCausality from to] }
 
 initialConcept :: (a -> InitialValue) -> Concept s e a
 initialConcept f = mempty { initial = f }
--- a/src/Tuura/Concept/Circuit/Basic.hs
+++ b/src/Tuura/Concept/Circuit/Basic.hs
@@ -1,4 +1,5 @@
 module Tuura.Concept.Circuit.Basic (
+    Causality (..),
     Interface (..),
     InitialValue (..),
     Invariant (..),
@@ -13,6 +14,8 @@ import Data.Monoid
 -- Abstract concepts
 -- * s is the type of states
 -- * e is the type of events
+
+data Causality e = AndCausality (e, e) | OrCausality ([e], e) deriving (Ord, Eq, Show)
 
 data Interface = Unused | Input | Output | Internal deriving (Ord, Eq, Show)
 
@@ -38,7 +41,7 @@ data Invariant e = NeverAll [e] deriving (Eq, Show)
 data Concept s e a = Concept
                    {
                        initial   :: a -> InitialValue,
-                       arcs      :: [([e], e)],
+                       arcs      :: [Causality e],
                        interface :: a -> Interface,
                        invariant :: [Invariant e]
                    }
@@ -55,10 +58,10 @@ instance Monoid (Concept s e a) where
                   }
 
 arcConcept :: e -> e -> Concept s e a
-arcConcept from to = mempty { arcs = [([from], to)] }
+arcConcept from to = mempty { arcs = [AndCausality (from, to)] }
 
 orCausality :: [e] -> e -> Concept s e a
-orCausality from to = mempty { arcs = [(from, to)] }
+orCausality from to = mempty { arcs = [OrCausality (from, to)] }
 
 initialConcept :: (a -> InitialValue) -> Concept s e a
 initialConcept f = mempty { initial = f }
--- a/src/Tuura/Concept/FSM/Translation.hs
+++ b/src/Tuura/Concept/FSM/Translation.hs
@@ -2,7 +2,8 @@ module Tuura.Concept.FSM.Translation where
 
 import Data.List
 import Data.List.Extra
-import Data.Ord (comparing)
+import qualified Data.List.NonEmpty as NonEmpty
+import Data.Ord
 import qualified Data.Set as Set
 import Control.Monad
 import Data.Char  (digitToInt)
@@ -84,7 +85,7 @@ translate circuit signs =
     case validateInitialState signs circuit of
       Valid -> do
           let allCause = addConsistency (arcs circuit) signs
-              sortedCause = concatMap handleArcs (groupSortOn snd (arcLists allCause))
+              sortedCause = concatMap handleArcs (NonEmpty.groupAllWith snd (arcLists allCause))
               initialState = getInitialState circuit signs
               allArcs = createAllArcs sortedCause
               reachables = findReachables allArcs initialState
@@ -96,7 +97,7 @@ translate circuit signs =
               internalSigns = filter ((==Internal) . interface circuit) signs
               reachableInvariants = filter (\i -> (encToInt i) `elem` reachables) invariants
               unreachables = ([0..2^(length signs) - 1] \\ invariantNos) \\ reachables
-          case validateFSM signs circuit reachableInvariants (invariant circuit) of
+          case (validateFSM signs reachableInvariants (invariant circuit)) <> (validateInterface signs circuit) of
               Valid -> do
                 let reachReport = genReachReport unreachables
                 genFSM inputSigns outputSigns internalSigns (map show reachableArcs) (show initialState) reachReport
@@ -113,21 +114,20 @@ fromBool :: Bool -> Tristate
 fromBool x = if x then triTrue else triFalse
 
 addConsistency :: Ord a => [Causality (Transition a)] -> [a] -> [Causality (Transition a)]
-addConsistency allArcs signs = nubOrd (allArcs ++ concatMap (\s -> [AndCausality (rise s, fall s), AndCausality (fall s, rise s)]) signs)
+addConsistency allArcs signs = nubOrd (allArcs ++ concatMap (\s -> [AndCausality (rise s) (fall s), AndCausality (fall s) (rise s)]) signs)
 
-handleArcs :: [([Transition a], Transition a)] -> [([Transition a], Transition a)]
+handleArcs :: NonEmpty.NonEmpty ([Transition a], Transition a) -> [([Transition a], Transition a)]
 handleArcs xs = map (\m -> (m, effect)) transCauses
         where
-            effect = snd (head xs)
-            effectCauses = map fst xs
+            effect = snd (NonEmpty.head xs)
+            effectCauses = NonEmpty.map fst xs
             transCauses = cartesianProduct effectCauses
 
-validateFSM :: Ord a => [a] -> CircuitConcept a -> [[Tristate]] -> [Invariant (Transition a)] -> ValidationResult a
-validateFSM signs circuit reachInvs invs
-    | invVio == [] && unused == [] = Valid
-    | otherwise = Invalid ((map UndefinedInitialState unused) ++ (map InvariantViolated invVio))
+validateFSM :: Ord a => [a] -> [[Tristate]] -> [Invariant (Transition a)] -> ValidationResult a
+validateFSM signs reachInvs invs
+    | invVio == [] = Valid
+    | otherwise = Invalid (map InvariantViolated invVio)
   where
-    unused  = filter ((==Undefined) . initial circuit) signs
     invsMapped = map (\(NeverAll is) -> (is, getInvariantStates (NeverAll is) signs)) invs
     invVio = nubOrd (map fst (concatMap (\i -> filter (\(_, x) -> i `elem` x) invsMapped) reachInvs))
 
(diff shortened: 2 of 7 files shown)


src/Tuura/Concept/Circuit/Basic.hs
```diff
@@ -14,6 +15,8 @@ import Data.Monoid
 -- * s is the type of states
 -- * e is the type of events
 
+data Causality e = AndCausality (e, e) | OrCausality ([e], e) deriving (Ord, Eq, Show)
```
snowleopard: Why use an extra pair here? I think `data Causality e = AndCausality e e | OrCausality [e] e` should lead to simpler code.

src/Tuura/Concept/FSM/Translation.hs
```diff
@@ -81,46 +81,55 @@ translateFSM circuitName ctype signs = do
...
+addConsistency :: Ord a => [Causality (Transition a)] -> [a] -> [Causality (Transition a)]
+addConsistency allArcs signs = nubOrd (allArcs ++ concatMap (\s -> [AndCausality (rise s, fall s), AndCausality (fall s, rise s)]) signs)
 
 handleArcs :: [([Transition a], Transition a)] -> [([Transition a], Transition a)]
-handleArcs arcLists = result
+handleArcs xs = map (\m -> (m, effect)) transCauses
         where
-            effect = snd (head arcLists)
-            effectCauses = map fst arcLists
+            effect = snd (head xs)
```
snowleopard: This will fail if `xs == []` -- is this expected? If no, let's fix this. If there is an assumption that the input list is non-empty, let's at least add a comment about this.

src/Tuura/Concept/FSM/Translation.hs
```diff
@@ -81,46 +81,55 @@ translateFSM circuitName ctype signs = do
...
+            effectCauses = map fst xs
             transCauses = cartesianProduct effectCauses
-            result = map (\m -> (m, effect)) transCauses
+
+validateFSM :: Ord a => [a] -> CircuitConcept a -> [[Tristate]] -> [Invariant (Transition a)] -> ValidationResult a
+validateFSM signs circuit reachInvs invs
+    | invVio == [] && unused == [] = Valid
+    | otherwise = Invalid ((map UndefinedInitialState unused) ++ (map InvariantViolated invVio))
```
snowleopard: No need for brackets around `map`. Function application (space) has highest precedence in Haskell.

src/Tuura/Concept/STG/Translation.hs
```diff
@@ -22,20 +22,20 @@ translate circuit signs =
...
-            "Error. \n" ++ addErrors errs
+        Invalid errs -> addErrors errs
 
 handleArcs :: Show a => [([Transition a], Transition a)] -> [String]
-handleArcs arcLists = addConsistencyTrans effect n ++ concatMap transition arcMap
+handleArcs xs = addConsistencyTrans effect n ++ concatMap transition arcMap
         where
-            effect = snd (head arcLists)
-            effectCauses = map fst arcLists
+            effect = snd (head xs)
```
snowleopard: Same here: `head xs` is a partial function (not defined on the empty list).

src/Tuura/Plato/Translation.hs
```diff
@@ -1,17 +1,25 @@
...
+    mappend (Invalid es) Valid = Invalid es
+    mappend Valid Valid = Valid
+    mappend (Invalid es) (Invalid fs) = Invalid (fs ++ es)
+    -- mappend (Defined x) (Defined y) = if x == y then Defined x else Inconsistent
+
 data ValidationError a = UnusedSignal a
                        | InconsistentInitialState a
                        | UndefinedInitialState a
-                       | InvariantViolated a
+                       | InvariantViolated [Transition a]
```
snowleopard, line 22: Shouldn't this be something like `InvariantViolated Invariant`?

src/Tuura/Plato/Translation.hs
```diff
@@ -27,41 +35,49 @@ instance Ord Signal
...
+             ++ unlines (map show invVio) ++ "\n"
+        else "")
     where
-        unused es = [ a | UnusedSignal a <- es ]
-        incons es = [ a | InconsistentInitialState a <- es ]
-        undefd es = [ a | UndefinedInitialState a <- es ]
+        unused = [ a | UnusedSignal a <- errs ]
+        incons = [ a | InconsistentInitialState a <- errs ]
+        undefd = [ a | UndefinedInitialState a <- errs ]
+        invVio = [ a | InvariantViolated a <- errs ]
```
snowleopard: Let's align right-hand sides of the above 4 lines.

src/Tuura/Plato/Translation.hs
```diff
@@ -27,41 +35,49 @@ instance Ord Signal
...
-validate signs circuit
-    | unused ++ inconsistent ++ undef == [] = Valid
-    | otherwise = Invalid ((map UnusedSignal unused) ++ (map InconsistentInitialState inconsistent)
-                  ++ (map UndefinedInitialState undef))
+validate signs circuit = (validateInitialState signs circuit) `mappend` (validateInterface signs circuit)
+
+validateInitialState :: Ord a => [a] -> CircuitConcept a -> ValidationResult a
+validateInitialState signs circuit
+    | undef ++ inconsistent == [] = Valid
+    | otherwise = Invalid ((map UndefinedInitialState undef) ++ (map InconsistentInitialState inconsistent))
```
snowleopard: Again, redundant brackets.

src/Tuura/Plato/Translation.hs
```diff
@@ -27,41 +35,49 @@ instance Ord Signal
...
+        incons = [ a | InconsistentInitialState a <- errs ]
+        undefd = [ a | UndefinedInitialState a <- errs ]
+        invVio = [ a | InvariantViolated a <- errs ]
 
 validate :: Ord a => [a] -> CircuitConcept a -> ValidationResult a
-validate signs circuit
-    | unused ++ inconsistent ++ undef == [] = Valid
-    | otherwise = Invalid ((map UnusedSignal unused) ++ (map InconsistentInitialState inconsistent)
-                  ++ (map UndefinedInitialState undef))
+validate signs circuit = (validateInitialState signs circuit) `mappend` (validateInterface signs circuit)
```
snowleopard: Again, redundant brackets.

snowleopard: And why not use `<>`?

snowleopard:
Thanks @jrbeaumont! See a bunch of comments above :)

src/Tuura/Concept/FSM/Translation.hs
jrbeaumont: It is expected that `xs` will never be empty, as the caller sorts based only on the existing list of arcs available in the `CircuitConcept`. I will add a comment here.

src/Tuura/Concept/STG/Translation.hs
jrbeaumont: This will also be non-empty. If you prefer, I can add an empty list clause just incase.

src/Tuura/Plato/Translation.hs
jrbeaumont: Yes, it should. I struggled with this as well and it's very simple haha

src/Tuura/Plato/Translation.hs
jrbeaumont: I wanted to use `mappend` so these are not confused with concepts, but `<>` would work and be neater so I will use it.

src/Tuura/Concept/FSM/Translation.hs
jrbeaumont: This throws errors for me.

src/Tuura/Concept/FSM/Translation.hs
snowleopard: @jrbeaumont Can't the list of arcs be empty? What if the specification is `circuit = empty`?

jrbeaumont: If it is, then on line 87, it will not have any arcs, which means that the list produced by `groupSortOn` will be empty, which means handle arcs will never be called as it is an empty list.

src/Tuura/Concept/STG/Translation.hs
snowleopard: I'm just worried whenever I see partial functions like `head` or `tail` and can't immediately see that the list can't be empty.

It's easy to write `head` but may be very hard to track down the cause of a runtime error!

src/Tuura/Plato/Translation.hs
jrbeaumont: This is true. I will add empty cases just in case.

src/Tuura/Concept/FSM/Translation.hs
snowleopard: Aha, I see, sounds plausible. Perhaps, we could try to use `Data.List.NonEmpty`?

https://hackage.haskell.org/package/semigroups-0.16/docs/Data-List-NonEmpty.html

Unfortunately, it doesn't have `groupSortOn`, but it does define `group` which returns `NonEmpty` list as a result, and provides a safe `head :: NonEmpty a -> a` :-) It may be a bit of an overkill, but is kind of cool.

jrbeaumont: This would help understanding of what's going on, so I will use this :)

src/Tuura/Plato/Translation.hs
jrbeaumont: I'm actually having a big problem getting `InvariantViolated (Invariant a)` to work. 

In the function `ValidateFSM` in `FSM.Translation`, line 125 onwards, if I change this to use `InvariantViolated (Invariant a)`, it throws errors which state that `a` is actually a `Transition a`, and if I change it to use this, then it throws errors stating it expects type `a`, but it's actually `Transition a`. 

I'm quite stuck with this. I will do the other changes, and push, and we can discuss this.

snowleopard: @jrbeaumont Sure, let's come back to this. In fact, you can postpone this change until a later PR if that makes things easier. (I'm pretty sure though that the proposed change will eventually work.)

jrbeaumont: I am sure it will work! It is just exceeding my knowledge as a Haskell programmer. 
I propose we postpone this until after this PR, and I will discuss it with you in person, so you can explain how I was doing it wrong haha.

jrbeaumont:
@snowleopard:
I have made these changes, including `NonEmpty`, excluding `InvariantViolated (Invariant a)`, as we discussed.

I have also moved some things around while I've been at it for use with Workcraft, as I have been working on updating Workcraft for use with the new features concurrently. The major thing that will be coming up is changing errors to be output to STDERR, and everything else to STDOUT. 

Have a look in your own time and let me know if you'd like any more changes made :)

src/Tuura/Concept/FSM/Translation.hs
```diff
@@ -81,46 +82,54 @@ translateFSM circuitName ctype signs = do
 
 translate :: (Show a, Ord a) => CircuitConcept a -> [a] -> String
 translate circuit signs =
-    case validate signs circuit of
+    case validateInitialState signs circuit of
       Valid -> do
           let allCause = addConsistency (arcs circuit) signs
-              sortedCause = concatMap handleArcs (groupSortOn snd allCause)
+              sortedCause = concatMap handleArcs (NonEmpty.groupAllWith snd (arcLists allCause))
```
snowleopard: Add `import Data.List.NonEmpty (NonEmpty, groupAllWith)` to imports and you can skip these qualifiers.

src/Tuura/Concept/FSM/Translation.hs
```diff
@@ -81,46 +82,54 @@ translateFSM circuitName ctype signs = do
...
-            else do
-              let reachReport = genReachReport unreachables
-              genFSM inputSigns outputSigns internalSigns (map show reachableArcs) (show initialState) reachReport
-      Invalid errs ->
-          "Error. \n" ++ addErrors errs
+              reachableInvariants = filter (\i -> (encToInt i) `elem` reachables) invariants
+              unreachables = ([0..2^(length signs) - 1] \\ invariantNos) \\ reachables
+          case (validateFSM signs reachableInvariants (invariant circuit)) <> (validateInterface signs circuit) of
+              Valid -> do
+                let reachReport = genReachReport unreachables
```
snowleopard: Stick to 4 spaces of indentation.

src/Tuura/Concept/STG/Translation.hs
```diff
@@ -22,20 +23,21 @@ translate circuit signs =
...
 
-handleArcs :: Show a => [([Transition a], Transition a)] -> [String]
-handleArcs arcLists = addConsistencyTrans effect n ++ concatMap transition arcMap
+-- Due to the caller, xs will never be empty, so `snd (head xs)` will never fail.
+handleArcs :: Show a => NonEmpty.NonEmpty ([Transition a], Transition a) -> [String]
+handleArcs xs = addConsistencyTrans effect n ++ concatMap transition arcMap
         where
-            effect = snd (head arcLists)
-            effectCauses = map fst arcLists
+            effect = snd (NonEmpty.head xs)
```
snowleopard, line 40: Qualification here is very useful, so keep it!

src/Tuura/Plato/Translation.hs
```diff
@@ -1,17 +1,26 @@
...
 
 import Tuura.Concept.Circuit.Basic
 import Tuura.Concept.Circuit.Derived
 
 data ValidationResult a = Valid | Invalid [ValidationError a] deriving Eq
 
+instance Monoid (ValidationResult a) where
+    mempty = mempty
+
+    mappend Valid (Invalid es) = Invalid es
```
snowleopard: You can simplify this as `mappend Valid x = x` etc.

jrbeaumont:
@snowleopard: I have fixed those now :)

src/Tuura/Plato/Translation.hs
```diff
@@ -12,8 +12,8 @@ data ValidationResult a = Valid | Invalid [ValidationError a] deriving Eq
 instance Monoid (ValidationResult a) where
     mempty = mempty
 
-    mappend Valid (Invalid es) = Invalid es
-    mappend (Invalid es) Valid = Invalid es
+    mappend Valid x = x
+    mappend x Valid = x
     mappend Valid Valid = Valid
```
snowleopard: This case is now redundant :)

snowleopard:
@jrbeaumont Thanks! One last touch (I'm surprised you didn't get a warning for redundant case).

jrbeaumont:
@snowleopard: I did get an error, I just missed it and was working on something else, my bad haha.

snowleopard:
Right. So, shall I merge this now?

jrbeaumont:
I would wait until the CI checks pass. I have tested it on my mac, but it would be nice to be sure it definitely passes on everything.

PR #75, opened 2017-05-16
Reconfigured causality  to use OR-causality as the base.

jrbeaumont:
All causality is now stated as `Causality [e] e` instead of being split into `AndCasuality` and `OrCasuality`. 

AND-causality will contain single element lists as the cause transitions, e.g: 
`rise a ~> rise c <> rise b ~> rise c` 
will be `Causality [a+] c+` and `Causality [b+] c+`

OR-causality will utilise this list format as previously, e.g.:
`[rise a, rise b] ~|~> rise c`
will be `Causality [a+, b+] c+`

Testing has proved this continues to work correctly.

### DIFF
--- a/TransTest.hs
+++ b/TransTest.hs
@@ -52,7 +52,7 @@ testHandshakeConcept = do
     assertEq (sort (arcs handshakeConcept)) (sort expected)
   where
     handshakeConcept = inputs[a] <> outputs [b] <> handshake00 a b
-    expected = [AndCausality (rise a) (rise b), AndCausality (rise b) (fall a), AndCausality (fall a) (fall b), AndCausality (fall b) (rise a)]
+    expected = [Causality [rise a] (rise b), Causality [rise b] (fall a), Causality [fall a] (fall b), Causality [fall b] (rise a)]
     [a, b] = map Signal [0 .. 1]
 
 testOrGateConcept :: IO ()
@@ -61,7 +61,7 @@ testOrGateConcept = do
     assertEq (sort (arcs orGateConcept)) (sort expected)
   where
     orGateConcept = inputs [a, b] <> outputs [c] <> initialise0 [a, b, c] <> orGate a b c
-    expected = [OrCausality [rise a, rise b] (rise c), AndCausality (fall a) (fall c), AndCausality (fall b) (fall c)]
+    expected = [Causality [rise a, rise b] (rise c), Causality [fall a] (fall c), Causality [fall b] (fall c)]
     [a, b, c] = map Signal [0..2]
 
 testTwoAndGates :: IO ()
@@ -70,8 +70,8 @@ testTwoAndGates = do
     assertEq (sort (arcs twoAndGatesConcept)) (sort expected)
   where
     twoAndGatesConcept = inputs [a, b, c, d] <> outputs [out] <> initialise0 [a, b, c, d, out] <> andGate a b out <> andGate c d out
-    expected = [AndCausality (rise a) (rise out), AndCausality (rise b) (rise out), AndCausality (rise c) (rise out),
-                AndCausality (rise d) (rise out), OrCausality [fall a, fall b] (fall out), OrCausality [fall c, fall d] (fall out)]
+    expected = [Causality [rise a] (rise out), Causality [rise b] (rise out), Causality [rise c] (rise out),
+                Causality [rise d] (rise out), Causality [fall a, fall b] (fall out), Causality [fall c, fall d] (fall out)]
     [a, b, c, d, out] = map Signal [0..4]
 
 testTwoDifferentCElements :: IO ()
--- a/TransTest.hs
+++ b/TransTest.hs
@@ -20,7 +20,8 @@ testDotGOutput = do
     putStrLn "=== testDotGOutput"
     assertEq threeSignalsConcept expected
   where
-    threeSignalsConcept = translate (inputs [a] <> outputs [b] <> internals [c] <> initialise0 [a, c] <> initialise1 [b]) signs
+    threeSignalsConcept = translate (inputs [a] <> outputs [b] <> internals [c]
+                       <> initialise0 [a, c] <> initialise1 [b]) signs
     a = Signal 0
     b = Signal 1
     c = Signal 2
@@ -52,7 +53,8 @@ testHandshakeConcept = do
     assertEq (sort (arcs handshakeConcept)) (sort expected)
   where
     handshakeConcept = inputs[a] <> outputs [b] <> handshake00 a b
-    expected = [Causality [rise a] (rise b), Causality [rise b] (fall a), Causality [fall a] (fall b), Causality [fall b] (rise a)]
+    expected = [Causality [rise a] (rise b), Causality [rise b] (fall a),
+                Causality [fall a] (fall b), Causality [fall b] (rise a)]
     [a, b] = map Signal [0 .. 1]
 
 testOrGateConcept :: IO ()
@@ -60,8 +62,10 @@ testOrGateConcept = do
     putStrLn "===testOrGateConcept"
     assertEq (sort (arcs orGateConcept)) (sort expected)
   where
-    orGateConcept = inputs [a, b] <> outputs [c] <> initialise0 [a, b, c] <> orGate a b c
-    expected = [Causality [rise a, rise b] (rise c), Causality [fall a] (fall c), Causality [fall b] (fall c)]
+    orGateConcept = inputs [a, b] <> outputs [c]
+                 <> initialise0 [a, b, c] <> orGate a b c
+    expected = [Causality [rise a, rise b] (rise c),
+                Causality [fall a] (fall c), Causality [fall b] (fall c)]
     [a, b, c] = map Signal [0..2]
 
 testTwoAndGates :: IO ()
@@ -69,18 +73,26 @@ testTwoAndGates = do
     putStrLn "===testTwoAndGates"
     assertEq (sort (arcs twoAndGatesConcept)) (sort expected)
   where
-    twoAndGatesConcept = inputs [a, b, c, d] <> outputs [out] <> initialise0 [a, b, c, d, out] <> andGate a b out <> andGate c d out
-    expected = [Causality [rise a] (rise out), Causality [rise b] (rise out), Causality [rise c] (rise out),
-                Causality [rise d] (rise out), Causality [fall a, fall b] (fall out), Causality [fall c, fall d] (fall out)]
+    twoAndGatesConcept = inputs [a, b, c, d] <> outputs [out]
+                      <> initialise0 [a, b, c, d, out] <> andGate a b out
+                      <> andGate c d out
+    expected = [Causality [rise a] (rise out), Causality [rise b] (rise out),
+                Causality [rise c] (rise out), Causality [rise d] (rise out),
+                Causality [fall a, fall b] (fall out),
+                Causality [fall c, fall d] (fall out)]
     [a, b, c, d, out] = map Signal [0..4]
 
 testTwoDifferentCElements :: IO ()
 testTwoDifferentCElements = do
     putStrLn "===testTwoDifferentCElements"
-    assertEq (sort (arcs cElementConcept)) (sort (arcs cElementConceptFromBuffers))
+    assertEq (sort (arcs cElementConcept))
+             (sort (arcs cElementConceptFromBuffers))
   where
-    cElementConcept = inputs [a, b] <> outputs [c] <> initialise0 [a, b, c] <> cElement a b c
-    cElementConceptFromBuffers = inputs [a, b] <> outputs [c] <> initialise0 [a, b, c] <> buffer a c <> buffer b c
+    cElementConcept = inputs [a, b] <> outputs [c]
+                   <> initialise0 [a, b, c] <> cElement a b c
+    cElementConceptFromBuffers = inputs [a, b] <> outputs [c]
+                              <> initialise0 [a, b, c] <> buffer a c
+                              <> buffer b c
     [a, b, c] = map Signal [0..2]
 
 assertEq :: (Eq a, Show a) => a -> a -> IO ()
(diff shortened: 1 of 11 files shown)


TransTest.hs
```diff
@@ -52,7 +52,7 @@ testHandshakeConcept = do
     assertEq (sort (arcs handshakeConcept)) (sort expected)
   where
     handshakeConcept = inputs[a] <> outputs [b] <> handshake00 a b
-    expected = [AndCausality (rise a) (rise b), AndCausality (rise b) (fall a), AndCausality (fall a) (fall b), AndCausality (fall b) (rise a)]
+    expected = [Causality [rise a] (rise b), Causality [rise b] (fall a), Causality [fall a] (fall b), Causality [fall b] (rise a)]
```
snowleopard: Let's also fit all lines to 80 characters if possible, so that it's easy to read in GitHub without scrolling to the right.

snowleopard:
Looks good! Shall I merge?

jrbeaumont:
Thank you :) 
I will go through and tidy up and make everything under 80 characters and merge it myself. 
I wanted your approval on the change, but there's no need to wait for me to tidy it. Thank you very much

snowleopard:
:+1:

By the way, I've just demoed Plato at Dialog -- everything went well :-)

jrbeaumont:
That's great to hear! 

As long as you didn't find any bugs then we're winning! Haha. 

I hope they have good and constructive feedback.

snowleopard:
The only issue is that for some reason images in your blog were distorted (incorrectly scaled). It was an older Windows machine with IE browser. I checked and everything is fine on my machine.

jrbeaumont:
I just checked on the standard IE on the university computers and that does not scale them correctly either. Every other browser I have tried does. 
I will look into this and see if there is an easy fix.

jrbeaumont:
@snowleopard: I'm gonna go through and tidy up all code so it is 80 characters or less, as in some cases I have found it might make some things clearer.
Is this OK? Would you mind reviewing before I merge?

src/Tuura/Concept/Circuit/Derived.hs
```diff
@@ -64,10 +64,14 @@ initialise :: Eq a => a -> Bool -> CircuitConcept a
 initialise a v = initialConcept $ \s -> if s == a then Defined v else Undefined
 
 initialise0 :: Eq a => [a] -> CircuitConcept a
-initialise0 as = if (as /= []) then initialise (head as) False <> initialise0 (tail as) else mempty
+initialise0 as = if (as /= [])
```
snowleopard: This should be a pattern matching on `as` instead if an if statement.

src/Tuura/Concept/FSM/Translation.hs
```diff
@@ -143,13 +171,20 @@ genFSM inputSigns outputSigns internalSigns arcStrs initialState reachReport =
 genReachReport :: (Eq a, Show a) => [a] -> String
 genReachReport es
```
snowleopard: Same here: use pattern matching.

src/Tuura/Concept/FSM/Translation.hs
```diff
@@ -158,9 +193,12 @@ fullListm :: ([TransitionX a], Transition a) -> [TransitionX a]
 fullListm (l,t) = (toTransitionX t):l
 
 -- Given [([a], b)], remove all b from a
-removeDupes :: Eq a => [([Transition a], Transition a)] -> [([Transition a], Transition a)]
+removeDupes :: Eq a => [([Transition a], Transition a)] ->
+                       [([Transition a], Transition a)]
 -- (filter ((/= ((signal . snd) x)) . signal) (fst x), snd x)
-removeDupes = map (ap ((,) . ap (filter . (. signal) . (/=) . signal . snd) fst) snd)
+removeDupes = map (ap
+              ((,) . ap (filter . (. signal) . (/=) . signal . snd) fst)
```
snowleopard: Whoa, that's impressive -- no idea what's going on here :-)

snowleopard: With great power comes great responsibility -- use with care :-)

snowleopard:
I've added some comments -- in general I feel you are starting to overuse point-free notation: the code is difficult to follow and may be hard to maintain :-)

src/Tuura/Concept/FSM/Translation.hs
jrbeaumont: Got to be honest here, this was Adrian's work. I'm just about sure what's going on here, but I might struggle to explain it.

jrbeaumont:
I will fix these, and then have a review of all the code, maybe trying to reduce the point-free notation, hopefully clearing some things up.

src/Tuura/Concept/FSM/Translation.hs
snowleopard: Ah, I didn't realise Adrian is so skilled in Haskell :-) With two new PhD students who are experienced Haskellers joining our team, we are getting dangerous ;)

jrbeaumont: Haha, we are indeed!

ARandomOWL: This one's my fault. I didn't figure how to implement the function in non-point-free form at the time :p @jrbeaumont Let me know if you need help reworking it.

jrbeaumont:
@snowleopard I have changed those. I will have a look through for some point-free notation, but if you spot anything, please let me know.

jrbeaumont:
Yeah, I am struggling to understand where and how to improve. If you could point to some problematic functions I will have a crack at clearing them up, seeing if I can find others in the process.

snowleopard:
@jrbeaumont Functions `removeDupes` and `addMissingSignals` could benefit from being rewritten in a clearer way, even if they will become longer as a result.

Note that in general if the line is getting too long, it is a good indication that perhaps some intermediate terms/functions should be defined in order to clarify the code. 

In many instances where you simply broken lines into multiple lines, I would instead consider coming up with a clear decomposition of the line into simple steps.

jrbeaumont:
@snowleopard: The latest commit to this branch will remove redundant arcs produced by `sequence`. Testing proves this works correctly, not affecting the previously working concepts, but using the example of an ORAND the STG produced one has two `c-` transition objects. 

Feel free to review and offer tips for improvements on the implementation. I tried to write these changes in a form that clear, let me know what you think.

This PR is not quite ready, I still need to tidy up the code you suggested earlier. I also forgot to try and shorten code in `Tuura.Plato.Translation` so I will fix this while I am at it. Should be done soon.

src/Tuura/Plato/Translation.hs
```diff
@@ -76,8 +78,30 @@ validateInterface signs circuit
...
+  where
+    prev n  = take n s
+    check x = checkForSupersets (s!!x) (prev x)
+    result  = map (\x -> if check x
+                        then []
+                        else s!!x
+                 )[0..(length s) - 1]
+
+checkForSupersets :: Eq a => [a] -> [[a]] -> Bool
+checkForSupersets current previous = any (`elem` subs) previous
```
snowleopard: Let's start by simplifying this function.

Can't you replace this with ``checkForSupersets current previous = any (`isSubsequenceOf` current) previous``?

snowleopard: Or, even shorter ``checkForSupersets current = any (`isSubsequenceOf` current)`` :) After which you probably no longer need a separate function.

jrbeaumont: My feelings with this that it was more efficient to run `subsequences` once for `current` as `isSubsequenceOf` will run it for every list in `previous`.

jrbeaumont: This would shorten it, and probably be nice and concise :) But would this affect the efficiency? It may be negligible in the grand scheme I suppose.

snowleopard: I think it will also be much faster, not slower!

* The current implementation needs to perform `elem` on each element of `x` in `previous`, which requires going through the list `subs` and comparing `x` with each element. If `subs` contains `m` elements, this will take `m * |x|` time in the worst case.

* My implementation needs to perform a single check `isSubsequenceOf` for each element of `x` in `previous`, which requires at most `|current| + |x|` time.

The latter is much faster, because `m` here contains all possible sublists of `current`, which means that `m = 2^|current|`!

snowleopard: Just try your function on `current = [1..20]` and `previous = [[100]]` -- I wonder how long it will take :)

jrbeaumont: That's fair. I think I just misunderstood the implementation of `isSubsequenceOf`. I just assumed it would run `subsequences` each time, then check all of these with something like `elem`. I thought if I made sure it only ran `subsequences` once, then it would be quicker. 

Running your example (with `current = [1..200]`) proves your method works almost instantaneously, whereas mine runs for a while, makes my computer very warm, and then the process kills itself after a while, so yours is most definitely better :)

snowleopard: Ah, I see where your confusion comes from. In the documentation it says that it is *equivalent* to `elem x (subsequences y)` and it is indeed equivalent in the sense that it gives the same result. But it's implemented a lot more efficiently than simply trying all possible subsequences.

src/Tuura/Plato/Translation.hs
```diff
@@ -76,8 +81,24 @@ validateInterface signs circuit
...
+
+removeSupersets :: Eq a => [[a]] -> [[a]]
+removeSupersets s = filter (not . null) result
+  where
+    prev n        = take n s
+    check current = any (`isSubsequenceOf` current)
+    result        = map (\x -> if check (s!!x) (prev x)
+                        then []
+                        else s!!x
+                 )[0..(length s) - 1]
```
snowleopard: This would be much nicer with list comprehensions:
```haskell
result = [ x | (x:xs) <- tails sortByLength, not (check x xs) ]
```
I guess you'll just need to reverse the sorting order so longer lists come first.

snowleopard: And you don't need any filter afterwards :-)

jrbeaumont: Good plan. 
There may be a function to order by size, largest first. I will have a look.

snowleopard: I guess something like `sortBy (comparing $ negate . length)` should do the trick.

jrbeaumont: Ill give it a try and see.

jrbeaumont: Works perfectly :) New version on the way.

jrbeaumont:
@snowleopard: I have cleared up everything I can. Some further revisions may be necessary on one or two functions but I need to discuss these with Adrian. 

Please let me know if you have any further suggestions :)

src/Tuura/Plato/Translation.hs
```diff
@@ -82,23 +82,20 @@ validateInterface signs circuit
...
+cartesianProduct l = removeSupersets sortAllLists
   where
     sequenced    = sequence (NonEmpty.toList l)
     removeDupes  = map nub sequenced
     removeNull   = filter (not . null) removeDupes
     sortAllLists = map sort removeNull
-    sortByLength = sortBy (comparing length) sortAllLists
 
 removeSupersets :: Eq a => [[a]] -> [[a]]
 removeSupersets s = filter (not . null) result
```
snowleopard: What's up with the `filter`? I think you've already removed all empty lists :)

src/Tuura/Plato/Translation.hs
```diff
@@ -82,23 +82,20 @@ validateInterface signs circuit
...
   where
     prev n        = take n s
```
snowleopard: The services of this guy are also no longer required.

jrbeaumont: Oh no I did remove these and then I reverted. 

I'm also working on the flipping thing we talked about, mixing branches :S

src/Tuura/Plato/Translation.hs
snowleopard: In fact, I don't think you need the other `filter` either.

src/Tuura/Plato/Translation.hs
jrbeaumont: These are now fixed and repushed :)

snowleopard: There is still `removeNull   = filter (not . null) removeDupes` which I think is not needed?

src/Tuura/Plato/Translation.hs
jrbeaumont: Sorry, only just spotted this second comment. 
This has also been fixed.

src/Tuura/Plato/Translation.hs
```diff
@@ -86,15 +86,12 @@ cartesianProduct l = removeSupersets sortAllLists
   where
     sequenced    = sequence (NonEmpty.toList l)
     removeDupes  = map nub sequenced
```
snowleopard: I'm pretty sure `removeSupersets` takes good care of duplicates.

jrbeaumont: If I remove this line, it goes back to having many redundancies in a translated STG.
It is there because, if you combine `[a-, c-], [a-,d-], [b-,c-], [b-,d-]`
Then you get results such as:
`[a-, a-, b-, b-], [a-, a-, b-, d-], [a-, a-, c-, b-]...`
This serves to remove the duplicates within each list

snowleopard: Ah, my bad. I didn't realise this does `nub` inside each list. Sure, it should be kept then.

jrbeaumont: No problem, there was a lot of unnecessary filtering and `nub`s :)

snowleopard:
@jrbeaumont Before merging this, could you add comments for each function? It may be difficult to understand what's going on in a few months from now.

jrbeaumont:
Sure thing. Each and every function, or just ones that are quite complicated?

snowleopard:
Well, if a function does something completely obvious then you can omit the comment (although in this case the comment is also easy to write and probably won't hurt).

jrbeaumont:
OK, I'll add some comments and you can have a quick review before merging :)

jrbeaumont:
@snowleopard I have added the comments into the `flip` branch in #76 so, so these functions include comments too. Please review them there.

snowleopard:
OK, I think we can now merge this one, right?

jrbeaumont:
Yes, please go ahead.

PR #77, opened 2017-06-19
Implemented bubble function to invert a chosen signal's transitions

jrbeaumont:
`bubble` is a concept transformation function which takes a signal and a concept and `toggle`s all of the transitions for that signal, be they cause/effect transitions in a causality, in AND/OR causalities, for the given concept. This acts as a negation on the chosen signal, be it an input or output.

This serves well to provide inverting gates, for example:
```haskell
bubble c (andGate a b c)
bubble a (orGate a b c)
bubble c (cElement a b c)
```
These will produce a NAND gate, an OR gate with one inverted input (`a`), and an inverting output C-element, respectively. 

This can also be used in combination, for multiple inversions, e.g:
```haskell
bubble b (bubble c (andGate a b c))
```
This will produce a NAND gate with one input (`b`) inverted.

### DIFF
--- a/src/Tuura/Concept/Circuit/Derived.hs
+++ b/src/Tuura/Concept/Circuit/Derived.hs
@@ -91,21 +91,23 @@ bubbleInitialValue s f y = if (y == s) then bubbleVal (f y) else f y
     bubbleVal (Defined v) = Defined (not v)
     bubbleVal x = x
 
+-- Perform the inversion for just transitions of the given signal
+invertSignal :: Eq a => a -> Transition a -> Transition a
+invertSignal a t
+    | signal t == a = toggle t
+    | otherwise     = t
+
 -- For the selected signal, toggle the directions of all transitions for this.
 bubbleCausality :: Eq a => a -> Causality (Transition a)
                         -> Causality (Transition a)
 bubbleCausality s (Causality f t)
-    | signal t == s = Causality f (toggle t)
-    | otherwise     = Causality (map invertCauses f) t
-  where
-    invertCauses c = if (signal c == s) then toggle c else c
+    | signal t == s = Causality (map (invertSignal s) f) (toggle t)
+    | otherwise     = Causality (map (invertSignal s) f) t
 
 -- Invert the invariant transition of the selected signal.
 bubbleInvariant :: Eq a => a -> Invariant (Transition a)
                         -> Invariant (Transition a)
-bubbleInvariant s (NeverAll es) = NeverAll (map invertInvars es)
-  where
-    invertInvars i = if (signal i == s) then toggle i else i
+bubbleInvariant s (NeverAll es) = NeverAll (map (invertSignal s) es)
 
 bubble :: Eq a => a -> CircuitConcept a -> CircuitConcept a
 bubble s c = mempty
@@ -115,6 +117,7 @@ bubble s c = mempty
                 interface = interface c,
                 invariant = fmap (bubbleInvariant s) (invariant c)
              }
+
 -- Signal-level concepts
 
 consistency :: CircuitConcept a
--- a/src/Tuura/Concept/Circuit/Derived.hs
+++ b/src/Tuura/Concept/Circuit/Derived.hs
@@ -88,18 +88,22 @@ bubble :: Eq a => a -> CircuitConcept a -> CircuitConcept a
 bubble s c = mempty
              {
                 initial = initial c,
-                arcs = map (bubbleBySignal s) (arcs c),
+                arcs = fmap (bubbleCausality s) (arcs c),
                 interface = interface c,
-                invariant = invariant c
+                invariant = fmap (bubbleInvariant s) (invariant c)
              }
-  where
 
-bubbleBySignal :: Eq a => a -> Causality (Transition a) -> Causality (Transition a)
-bubbleBySignal s (Causality f t)
+bubbleCausality :: Eq a => a -> Causality (Transition a) -> Causality (Transition a)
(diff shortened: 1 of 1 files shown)


snowleopard:
Nice! But I think we should also toggle corresponding initial state bit, as well as the invariant.

jrbeaumont:
I thought about this, and I agree with invariants in the chosen concept to bubble, but my issue with initial state is that you would have to include the whole concept, i.e. More than just the gate/protocol/concept you wish to bubble. This means that if you want to simply have a NAND gate, but a non-inverting C-element later on using the output of the NAND, it can become confusing, and you toggle every causality in the system.

snowleopard:
Not sure I understand, can you give an example?

I don't see how the following doesn't work:
```haskell
nAndGate a b c = bubble c $ andGate a b c

test a b c = bubble c (nAndGate a b c <> initialise0 [a, b, c])
```

jrbeaumont:
Well I have just had a thought, and we can toggle all initial states and invariants in the given concept, which can be:

```haskell
nAndGate a b c = bubble c $ andGate a b c
```

this will just toggle the causalities for c in the AND gate. 
Or we can use it as:

```haskell
wholeSystem x y z = bubble z $ operation x y z <> initialise0 [x, y, z] <> never [rise y, rise z]
```
which will invert everything that z does.

snowleopard:
Yes, inverting everything is the idea: it's the same as with `dual` but only for one signal.

jrbeaumont:
I will implement it now.

jrbeaumont:
I have implemented the invariant toggles, but I cannot find a way to invert the initial state for the given signal. 
The implementation for dual can do it regardless of signal, because it does all initial states.
For bubble, I try to use the signal to find the initial state for just this signal, but it does not work, instead throwing compilation errors however I try to do it. 
I also need to keep the initial states that are not for the given signal as they are.

snowleopard:
Well, `invariant` is a function that takes a signal as the input. Try to go this way:
```haskell
bubbleInvariant :: Eq a => a -> (a -> InitialValue) -> (a -> InitialValue)
bubbleInvariant x f y = ...
```

jrbeaumont:
@snowleopard There we go! All works well :)

src/Tuura/Concept/Circuit/Derived.hs
```diff
@@ -82,6 +84,39 @@ dual c = mempty
...
+    bubbleVal x = x
+
+-- For the selected signal, toggle the directions of all transitions for this.
+bubbleCausality :: Eq a => a -> Causality (Transition a)
+                        -> Causality (Transition a)
+bubbleCausality s (Causality f t)
+    | signal t == s = Causality f (toggle t)
+    | otherwise     = Causality (map invertCauses f) t
+  where
+    invertCauses c = if (signal c == s) then toggle c else c
```
snowleopard: You also use this function below. Maybe define it as a top-level function `toggleSignal`?

src/Tuura/Concept/Circuit/Derived.hs
```diff
@@ -82,6 +84,39 @@ dual c = mempty
...
+bubbleInitialValue s f y = if (y == s) then bubbleVal (f y) else f y
+  where
+    bubbleVal (Defined v) = Defined (not v)
+    bubbleVal x = x
+
+-- For the selected signal, toggle the directions of all transitions for this.
+bubbleCausality :: Eq a => a -> Causality (Transition a)
+                        -> Causality (Transition a)
+bubbleCausality s (Causality f t)
+    | signal t == s = Causality f (toggle t)
```
snowleopard: Shouldn't you do `(map invertCauses f)` here too?

snowleopard:
@jrbeaumont Nice! I've added a couple of comments.

src/Tuura/Concept/Circuit/Derived.hs
jrbeaumont: I suppose it should, but is not often there are causes for the same signal as the effect.

snowleopard: Well, `a+ -> a-` should still be treated correctly even if it's not very meaningful :)

jrbeaumont: I suppose the inversion of consistency is the same, so it can't hurt :)

jrbeaumont:
@snowleopard Fixed those issues :)

src/Tuura/Concept/Circuit/Derived.hs
```diff
@@ -82,6 +84,42 @@ dual c = mempty
...
 
+-- Give the opposite initial state for the chosen signal
+bubbleInitialValue :: Eq a => a -> (a -> InitialValue) -> (a -> InitialValue)
+bubbleInitialValue s f y = if (y == s) then bubbleVal (f y) else f y
+  where
+    bubbleVal (Defined v) = Defined (not v)
+    bubbleVal x = x
+
+-- Perform the inversion for just transitions of the given signal
+invertSignal :: Eq a => a -> Transition a -> Transition a
```
snowleopard: There is function `toggle` which is almost the same, so for consistency we should keep `toggle` in the name instead of introducing additional `invert` naming. I think we need something like `toggleSignal`, `toggleSpecific` or similar.

jrbeaumont: `toggleSpecific` works for me.

jrbeaumont:
Rename done.

snowleopard:
Thanks! Now merged.

PR #78, opened 2017-07-18
Added XOR gate and function to remove redundant combinations of cause transitions

jrbeaumont:
Redundancy removal been tested with the XOR gate in question, a 3 input XOR, and some older examples, such as an AO22, to ensure this has not affected the existing working.

### DIFF
--- a/src/Tuura/Plato/Translation.hs
+++ b/src/Tuura/Plato/Translation.hs
@@ -90,12 +90,10 @@ validateInterface signs circuit
 -- Perform cartesian product on list of lists. This will also sort and remove
 -- duplicates in sublists, and remove supersets for the most compact form.
 cartesianProduct :: Ord a => NonEmpty.NonEmpty [Transition a] -> [[Transition a]]
-cartesianProduct l = removeSupersets sortAllLists
+cartesianProduct l = removeSupersets $ removeRedundancies $ 
+                       map (sort . nub) sequenced
   where
     sequenced    = sequence (NonEmpty.toList l)
-    removeDupes  = map nub sequenced
-    removeRedun  = removeRedundancies removeDupes
-    sortAllLists = map sort removeRedun
 
 -- Sort list of lists from largest length to shortest, then remove any lists
 -- that have shorter subsequences within the rest of the list.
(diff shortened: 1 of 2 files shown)


src/Tuura/Plato/Translation.hs
```diff
@@ -89,21 +89,27 @@ validateInterface signs circuit
...
 -- that have shorter subsequences within the rest of the list.
-removeSupersets :: Eq a => [[a]] -> [[a]]
+removeSupersets :: Eq a => [[Transition a]] -> [[Transition a]]
 removeSupersets s = [ x | (x:xs) <- tails sortByLength, not (check x xs) ]
   where
     check current = any (`isSubsequenceOf` current)
     sortByLength  = sortBy (comparing $ negate . length) s
 
+removeRedundancies :: Eq a => [[Transition a]] -> [[Transition a]]
+removeRedundancies s = filter (\ts ->
```
snowleopard: You can drop `\ts -> ` and ` ts` here. And then the same with ` s`. (This is called eta reduction/conversion: https://wiki.haskell.org/Eta_conversion).

src/Tuura/Plato/Translation.hs
```diff
@@ -89,21 +89,27 @@ validateInterface signs circuit
 
 -- Perform cartesian product on list of lists. This will also sort and remove
 -- duplicates in sublists, and remove supersets for the most compact form.
-cartesianProduct :: Ord a => NonEmpty.NonEmpty [a] -> [[a]]
+cartesianProduct :: Ord a => NonEmpty.NonEmpty [Transition a] -> [[Transition a]]
 cartesianProduct l = removeSupersets sortAllLists
```
snowleopard: I would simplify this as `removeSupersets $ map (sort . removeRedundancies . nub) sequenced`.

snowleopard:
Thanks, looks good! Just a couple of minor comments.

src/Tuura/Plato/Translation.hs
jrbeaumont: Removing `s` is fine. Removing `ts` is not. 

```haskell
filter (all (\t -> not ((toggle t) `elem`)))
```
throws the error

```   
 * Couldn't match expected type `Bool'
                  with actual type `t0 (Transition a) -> Bool'
    * In the first argument of `not', namely `((toggle t) `elem`)'
      In the expression: not ((toggle t) `elem`)
      In the first argument of `all', namely
        `(\ t -> not ((toggle t) `elem`))'
    * Relevant bindings include
        t :: Transition a (bound at src\Tuura\Plato\Translation.hs:109:36)
        removeRedundancies :: [[Transition a]] -> [[Transition a]]
          (bound at src\Tuura\Plato\Translation.hs:109:1)
```

jrbeaumont:
@snowleopard: I have tidied those functions the best I can. See a reply to your comments above, but otherwise this still works correctly.

snowleopard:
Thanks, merged!

I didn't notice that ``\ts -> all (\t -> not ((toggle t) `elem` ts)) ts`` refers to `ts` twice, so yep, we can't get rid of it unfortunately :)

PR #80, opened 2017-07-30
Bool to concepts

jrbeaumont:
Added a tool to convert Boolean set/reset functions into a concept specification.

This tool generates CNF forms of the functions from whatever format they are input.
It then simplifies these. CNF can be used to generate a concept specification.

Also moved CNF -> DNF converter (Cartesian Product) and DNF simplifier into `BooleanFunctions`, so operations on Boolean Functions, CNF or DNF form are together.

Also, on a minor note, the `boolToConcepts` tool will allow a user to enter a set/reset function with `-s`/`-r` flags respectively. `boolToConcepts` and `translate` tools will both allow a user to enter an output file path using the `-o` flag.


snowleopard:
Let's start with modules. I think it is useful to move all Boolean stuff into `Tuura.Boolean.*`:
* `Tuura.Parser.Boolean` -> `Tuura.Boolean.Parser`.
* `Tuura.Plato.BoolToConcept.BooleanFunctions` -> `Tuura.Boolean`.

However, some things should still be left in the tool:
* `fromFunctions :: String -> String -> (Bool, String)`.
* `createConceptSpec :: [String] -> CNF String -> CNF String -> String`.

I think some generic parts of the above two functions can be moved into `Tuura.Concept.*` where you could allow users to create concepts directly in Haskell from set/reset functions. For example:
```haskell
combinationalGate :: Expr a -> a -> CircuitConcept a
combinationalGate setFunction signal = ...
```

Then there will be nice equivalences, e.g. `andGate a b c == combinationalGate (and (Var a) (Var b)) c`.

jrbeaumont:
These sound like a good idea. 
Do you mean`fromFunctions` and `createConceptSpec` should be left in `Tuura.Boolean`?
Or keep a module called `Tuura.Plato.BoolToConcept`? And leave these in there?

jrbeaumont:
Also, as a thought, it will be tough to take an `Expr a` and with the resulting concept specification, and run `translate/Main.hs`. I am not sure how possible it is to run a translation from within a translation, if you know what I mean.

jrbeaumont:
Ignore the previous comment, I've worked out how to produce a `CircuitConcept a` from `Expr a`

jrbeaumont:
An issue I have found with `combinationalGate :: Expr a -> a -> CircuitConcept a` is that several
`Tuura.Boolean` functions require `Ord a =>` and so, using it in a concept specification doesn't work without being able to add `Ord a =>` in the declaration.

jrbeaumont:
Refactored the modules.

src/Tuura/Boolean/Options.hs
```diff
@@ -1,4 +1,4 @@
-module Tuura.Plato.BoolToConcept.Options (Options(..), getOptions) where
+module Tuura.Boolean.Options (Options(..), getOptions) where
```
snowleopard: Wait, this module is very specific to the executable, it should stay in `Plato`.

snowleopard:
> using it in a concept specification doesn't work without being able to add `Ord a =>` in the declaration.

Ah, sure, go ahead and add it -- it's a relatively minor constraint.

src/Tuura/Boolean/Options.hs
jrbeaumont: There is separate `Options`, one for `translate/Main.hs` and one for `boolToConcepts/Main.hs`. See the root directory.

jrbeaumont:
It needs to be added in the declaration in a Concept file. e.g. 
```Haskell
circuit :: Ord a => a -> a -> a -> CircuitConcept a
circuit a b c = ...
```

src/Tuura/Boolean/Options.hs
snowleopard: I'm talking about the module `Tuura.Boolean.Options`. It's not generally useful (noone will ever use it apart from `BoolToConcept/Main`), so it shouldn't live in `Tuura.Boolean.*`. It should be located in the executable file hierarchy.

snowleopard:
> It needs to be added in the declaration in a Concept file. e.g.

Hmm, don't do this yet. Just define the `combinationalGate` function if you can.

src/Tuura/Boolean/Options.hs
jrbeaumont: Ahh right. I'll move it.

jrbeaumont:
I will see what I can do, maybe provide some errors if there are any and we can work it out from there. 
I was thinking of maybe providing one for a `String` too. I.e
`combinationGate :: String -> Signal -> CircuitConcept a`
Where the String is then parsed for a Boolean tree. 
So for example `a*b` would also work. What do you think?

jrbeaumont:
I have implemented `combinationalGate`, and adjusted some of the functions to avoid needing `ord`, and it can be used in concept specifications now.

src/Tuura/Boolean.hs
```diff
@@ -0,0 +1,86 @@
...
+import Tuura.Boolean.Parser
+import Data.List
+import Data.List.Extra
+import Data.Foldable
+import Data.Maybe
+import Data.Ord
+
+type CNF a = [[Literal a]]
+
+type DNF a = [[Literal a]]
```
snowleopard: I'm thinking, perhaps, it's better to wrap these into newtypes, so it's not possible to accidentally use one instead of the other. Something like `newtype CNF a = CNF { fromCNF :: [[Literal a]] }`.

src/Tuura/Boolean.hs
```diff
@@ -0,0 +1,86 @@
...
+
+data Literal a = Literal { variable :: a, polarity :: Bool } deriving (Eq, Ord)
+
+convertToCNF :: Eq a => (Expr a) -> CNF a
```
snowleopard: Drop redundant parens: `Expr a` works fine.

src/Tuura/Boolean.hs
```diff
@@ -0,0 +1,86 @@
...
+convertToCNF expr = cnf
+  where
+    vars = toList expr
+    values = mapM (const [False, True]) vars
+    fs = filter (\v -> not $ eval expr (\x -> getValue x vars v)) values
+    genVar val var = Literal var val
```
snowleopard: Why do you need `genVar`? You can use `Literal` instead.

src/Tuura/Boolean.hs
```diff
@@ -0,0 +1,86 @@
...
+type DNF a = [[Literal a]]
+
+data Literal a = Literal { variable :: a, polarity :: Bool } deriving (Eq, Ord)
+
+convertToCNF :: Eq a => (Expr a) -> CNF a
+convertToCNF expr = cnf
+  where
+    vars = toList expr
+    values = mapM (const [False, True]) vars
+    fs = filter (\v -> not $ eval expr (\x -> getValue x vars v)) values
```
snowleopard: Looks like `getValue` should be defined locally here, it's not needed anywhere else.

snowleopard: Also, flip its arguments, so you can write simply `not $ eval expr (getValue vars v)` here.

src/Tuura/Boolean.hs
```diff
@@ -0,0 +1,86 @@
...
+type CNF a = [[Literal a]]
+
+type DNF a = [[Literal a]]
+
+data Literal a = Literal { variable :: a, polarity :: Bool } deriving (Eq, Ord)
+
+convertToCNF :: Eq a => (Expr a) -> CNF a
+convertToCNF expr = cnf
+  where
+    vars = toList expr
```
snowleopard: I think you need `vars = nub $ toList expr` here?

src/Tuura/Boolean.hs
```diff
@@ -0,0 +1,86 @@
...
+    values = mapM (const [False, True]) vars
+    fs = filter (\v -> not $ eval expr (\x -> getValue x vars v)) values
+    genVar val var = Literal var val
+    sim = simplifyCNF $ map (\v -> nub $ (map (\f -> (genVar (v !! f) (vars !! f))) [0..(length vars - 1)])) fs
```
snowleopard: I think you shouldn't do the simplification as part of conversion to CNF. Let's separate these two concerns.

src/Tuura/Boolean.hs
```diff
@@ -0,0 +1,86 @@
...
+
+genConcepts :: Bool -> [Literal String] -> String
+genConcepts v e
+    | (length e) == 1 = causes ++ op ++ effect
+    | otherwise       = "[" ++ causes ++ "]" ++ op ++ effect
+  where
+    causes = unwords $ intersperse "," $ map (direction) e
+    direction (Literal a True)  = "rise " ++ a
+    direction (Literal a False) = "fall " ++ a
+    op       = if (length e > 1) then " ~|~> " else " ~> "
```
snowleopard: This seems to be useless -- just use `" ~|~> "` and `" ~> "` directly above.

src/Tuura/Boolean.hs
```diff
@@ -0,0 +1,86 @@
...
+    vars = toList expr
+    values = mapM (const [False, True]) vars
+    fs = filter (\v -> not $ eval expr (\x -> getValue x vars v)) values
+    genVar val var = Literal var val
+    sim = simplifyCNF $ map (\v -> nub $ (map (\f -> (genVar (v !! f) (vars !! f))) [0..(length vars - 1)])) fs
+    cnf = map (map (\s -> Literal (variable s) (not $ polarity s))) sim
+
+genConcepts :: Bool -> [Literal String] -> String
+genConcepts v e
+    | (length e) == 1 = causes ++ op ++ effect
```
snowleopard: You don't need parentheses here. Perhaps, run your code through `hlint` -- it will catch a lot of small things like this.

src/Tuura/Boolean.hs
```diff
@@ -0,0 +1,86 @@
...
+convertToCNF expr = cnf
+  where
+    vars = toList expr
+    values = mapM (const [False, True]) vars
+    fs = filter (\v -> not $ eval expr (\x -> getValue x vars v)) values
+    genVar val var = Literal var val
+    sim = simplifyCNF $ map (\v -> nub $ (map (\f -> (genVar (v !! f) (vars !! f))) [0..(length vars - 1)])) fs
+    cnf = map (map (\s -> Literal (variable s) (not $ polarity s))) sim
+
+genConcepts :: Bool -> [Literal String] -> String
```
snowleopard: The role of the `Bool` is unclear. Add docs.

src/Tuura/Boolean.hs
jrbeaumont: I didn't realise this was a thing :S I will run it, and see what else it picks up :)

src/Tuura/Boolean.hs
```diff
@@ -0,0 +1,86 @@
...
+genConcepts v e
+    | (length e) == 1 = causes ++ op ++ effect
+    | otherwise       = "[" ++ causes ++ "]" ++ op ++ effect
+  where
+    causes = unwords $ intersperse "," $ map (direction) e
+    direction (Literal a True)  = "rise " ++ a
+    direction (Literal a False) = "fall " ++ a
+    op       = if (length e > 1) then " ~|~> " else " ~> "
+    effect   = if (v) then "rise out" else "fall out"
```
snowleopard: Why is it fixed to `out` name? I think making it general may be useful.

snowleopard:
I've left a few comments, will do another review later today.

src/Tuura/Boolean.hs
jrbeaumont: This was just for generating concepts from functions, so a user does not have to specify when they provide just the set and reset functions. I can make it general and use `out` if there is no name given.

jrbeaumont:
I have made those changes. `hlint` really helped.

src/Tuura/Boolean.hs
```diff
@@ -11,76 +11,73 @@ import Data.Foldable
...
   where
     check current = any (`isSubsequenceOf` current)
     sortByLength  = sortBy (comparing $ negate . length) s
 
-removeRedundancies :: Eq a => CNF a -> CNF a
-removeRedundancies = filter (\ts -> all (\t -> not ((neg t) `elem` ts)) ts)
+removeRedundancies :: Eq a => [[Literal a]] -> [[Literal a]]
+removeRedundancies = filter (\ts -> all (\t -> neg t `notElem` ts) ts)
   where
     neg x = Literal { variable = variable x, polarity = not $ polarity x}
```
snowleopard: I think this is a bit clearer: `neg x = Literal (variable x) (not $ polarity x)`.

snowleopard: Also, this looks like a generally useful function -- you need it above. Perhaps, call it `invert`?

src/Tuura/Boolean.hs
```diff
@@ -11,76 +11,73 @@ import Data.Foldable
...
-    vars = toList expr
+    vars = nub $ toList expr
     values = mapM (const [False, True]) vars
-    fs = filter (\v -> not $ eval expr (\x -> getValue x vars v)) values
-    genVar val var = Literal var val
-    sim = simplifyCNF $ map (\v -> nub $ (map (\f -> (genVar (v !! f) (vars !! f))) [0..(length vars - 1)])) fs
-    cnf = map (map (\s -> Literal (variable s) (not $ polarity s))) sim
+    fs = filter (not . eval expr . getValue vars) values
+    sim = CNF $ map (\v -> nub $ map (\f -> Literal (vars !! f) (v !! f)) [0..(length vars - 1)]) fs
+    cnf = CNF $ map (map (\s -> Literal (variable s) (not $ polarity s))) (fromCNF sim)
```
snowleopard: Here you are also inverting a literal.

src/Tuura/Boolean.hs
```diff
@@ -11,76 +11,73 @@ import Data.Foldable
...
+    getValue vs vals v = fromJust $ lookup v $ zip vs vals
 
-genConcepts :: Bool -> [Literal String] -> String
-genConcepts v e
-    | (length e) == 1 = causes ++ op ++ effect
-    | otherwise       = "[" ++ causes ++ "]" ++ op ++ effect
+-- Generates a concept for output signal to rise if v = True, fall if v = False
+genConcepts :: Bool -> String -> [Literal String] -> String
```
snowleopard: Could you pass `Transition String` instead of `Bool` and `String`? I think conceptually this is what you want.

jrbeaumont: Actually I think so, and there is a `show` for it too I beleive. If not, I may add it anwyay.

src/Tuura/Boolean.hs
```diff
@@ -11,76 +11,73 @@ import Data.Foldable
...
+simplifyDNF x = DNF (removeRedundancies $ removeSupersets $ fromDNF x)
 
 simplifyCNF :: Eq a => CNF a -> CNF a
-simplifyCNF c = removeSupersets $ removeCancels vars c
+simplifyCNF c = CNF (removeSupersets $ fromCNF $ removeCancels vars c)
   where
-    vars = nub $ concatMap (map variable) c
+    vars = nub $ concatMap (map variable) (fromCNF c)
 
 removeCancels :: Eq a => [a] -> CNF a -> CNF a
```
snowleopard: Some documentation/examples would help here.

src/Tuura/Boolean.hs
```diff
@@ -11,76 +11,73 @@ import Data.Foldable
...
-removeSupersets :: Eq a => CNF a -> CNF a
+removeSupersets :: Eq a => [[Literal a]] -> [[Literal a]]
 removeSupersets s = [ x | (x:xs) <- tails sortByLength, not (check x xs) ]
   where
     check current = any (`isSubsequenceOf` current)
     sortByLength  = sortBy (comparing $ negate . length) s
 
-removeRedundancies :: Eq a => CNF a -> CNF a
-removeRedundancies = filter (\ts -> all (\t -> not ((neg t) `elem` ts)) ts)
+removeRedundancies :: Eq a => [[Literal a]] -> [[Literal a]]
```
snowleopard: I think all these functions need docs, as they are not trivial and their type signatures don't really help to understand what they are doing.

src/Tuura/Boolean/Parser.hs
```diff
@@ -24,17 +24,17 @@ instance Show a => Show (Expr a) where
...
-not :: (Expr a) -> (Expr a)
+not :: Expr a -> Expr a
 not (Not e)     = e
 not (And e1 e2) = Or (not e1) (not e2)
 not (Or e1 e2)  = And (not e1) (not e2)
 not (SubExpr e) = not e
 not (Var c)     = Not (Var c)
 
-or :: (Expr a) -> (Expr a) -> (Expr a)
+or :: Expr a -> Expr a -> Expr a
```
snowleopard: I'm confused: why do you need `or`? Can't you use `Or` instead?

src/Tuura/Boolean/Parser.hs
```diff
@@ -24,17 +24,17 @@ instance Show a => Show (Expr a) where
   show (Var c)     = show c
   show (SubExpr e) = "(" ++ show e ++ ")"
 
-not :: (Expr a) -> (Expr a)
+not :: Expr a -> Expr a
```
snowleopard: I'm confused: why do you need `not`? Can't you use `Not` instead?

src/Tuura/Boolean/Parser.hs
jrbeaumont: Or is less useful now we have conversion to CNF. I will remove it.

src/Tuura/Boolean/Parser.hs
jrbeaumont: `not` is used for simply providing inversions in a nicer way. It can most likely be removed, and if it causes an issues, that is in problems in other functions.

src/Tuura/Boolean/Parser.hs
snowleopard: I think you misunderstood me: in your parser, you can just use `Or` instead of `or` and everything will work. You do this with `And` already.

snowleopard:
OK, that's enough comments for now -- please ping me when for the next iteration.

src/Tuura/Boolean/Parser.hs
jrbeaumont: Sorry, I did understand, it was just auto-capitalized :S
The `or` function is now useless.

src/Tuura/Boolean.hs
jrbeaumont: Update: Show will not work, because the type of `Transition a` has the declaration 
```haskell
show (Transition s True) = "rise " ++ show s
```
and similar for a falling transition.
This causes the `String` value that `s` is in this case to be surrounded by `"` characters.

snowleopard: Well, it's OK to write a custom pretty-printer. The important thing is the interface is going to be clearer with `Transition`.

jrbeaumont: `Transition` will make it clearer, and I have done this.
Instead of just doing `show Transition String`, I just use the value of the transition to determine 
`rise` or `fall`.

jrbeaumont:
@snowleopard: Next iteration done. 
I have noticed that not using parentheses in a set function causes it to evaluate it incorrectly. I should probably try to fix this. I believe it was initially my intention with `or` to try and force parentheses around every sub expression which would stop the ordering it takes. 
`not` was just there for clearer outputs with `show` so can continue to remain removed.

snowleopard:
> I have noticed that not using parentheses in a set function causes it to evaluate it incorrectly.

Can you give an example?

jrbeaumont:
`a*!b+!a*b`
An XOR, but it interprets this as `a*!b`

snowleopard:
Is it parsed as `a*(!b + !a*b)`? Can you show the `Expr` after parsing?

jrbeaumont:
After parsing it is `"a" * !"b" + !"a" * "b"`. It doesn't help haha

snowleopard:
I think you should derive `Show` for Expr, as otherwise it's not very useful. What does it print if derived?

jrbeaumont:
`Show` is derived for `Expr`. The issue is I think it parses it too ambiguously, so `eval` doesn't work correctly. 
Showing after `convertToCNF` reveals: ` !"a"+!"b" * "a"+!"b" * "a"+"b"`

jrbeaumont:
Here is the way it has parsed it: `AND (OR (AND ("a"NOT"b" ) NOT"a" ) "b") `
So `(((a * !b) + !a) *b)`

snowleopard:
> `Show` is derived for `Expr`.

No, it is not. It's your custom implementation, I meant `deriving Show`.

snowleopard:
> So `(((a * !b) + !a) *b)`

Aha, I see. So the parser doesn't handle operator precedence correctly. You need to fix this (and add some tests).

jrbeaumont:
I think I've already fixed the precedence haha. 
New result: `Or (And (Var "a") (Not (Var "b"))) (And (Var "b") (Not (Var "a")))`
Which is: ` a*!b + b*!a`

jrbeaumont:
The precedence is now on `And`. So if someone wants to do an AO22, then they can do:
`a*b+c*d`.
However, if someone wants to do an OA22, the need to add parentheses:
`(a+b)*(c+d)`
What do you think?

snowleopard:
@jrbeaumont Yes, these results look fine! Could you show the commit that does the fix?

jrbeaumont:
@snowleopard: The commit is pushed

snowleopard:
Oh, that's a tiny change :)

jrbeaumont:
More than a full-stop ;)

jrbeaumont:
I'll add some testing for `boolToConcept` next.

jrbeaumont:
@snowleopard Testing implemented and working.

src/Tuura/Boolean.hs
```diff
@@ -0,0 +1,96 @@
...
+
+data Literal a = Literal { variable :: a, polarity :: Bool } deriving (Eq, Ord, Show)
+
+convertToCNF :: Eq a => Expr a -> CNF a
+convertToCNF expr = cnf
+  where
+    vars = nub $ toList expr
+    values = mapM (const [False, True]) vars
+    fs = filter (not . eval expr . getValue vars) values
+    sim = CNF $ map (\v -> nub $ map (\f -> Literal (vars !! f) (v !! f)) [0..(length vars - 1)]) fs
```
snowleopard: I think this can be simplified:
* `nub` is probably redundant?
* By inserting `not` before `v !! f` you immediately get the `cnf` -- no need to invert `sim` in the next line.

src/Tuura/Boolean.hs
```diff
@@ -0,0 +1,96 @@
...
+    cnf = CNF $ map (map invert) (fromCNF sim)
+    getValue vs vals v = fromJust $ lookup v $ zip vs vals
+
+-- Generates a concept for output signal to rise if v = True, fall if v = False
```
snowleopard: Comment seems to be out-of-date

src/Tuura/Boolean.hs
```diff
@@ -0,0 +1,96 @@
...
+    var = Literal v True
+    nVar = Literal v False
+    relevant = filter (var `elem`) (fromCNF whole)
+    nRelevant = map (replace [var] [nVar]) relevant
+    existInWhole = filter (`elem` fromCNF whole) nRelevant
+    toBeRemoved = map (replace [nVar] [var]) existInWhole
+    replacements = map (delete var) toBeRemoved
+    newWhole = CNF (fromCNF whole ++ replacements)
+
+-- Apply cartesian produce to a CNF function to get DNF,
```
snowleopard: "produce" -> "product".

src/Tuura/Concept/Circuit/Derived.hs
```diff
@@ -199,3 +201,14 @@ internals ints = interfaceConcept $ \s ->
                  if s `elem` ints
                  then Internal
                  else Unused
+
+combinationalGate :: Eq a => Expr a -> a -> CircuitConcept a
```
snowleopard: What about also adding gates where both set and reset functions are specified? I'm not sure about the best name thought. The idea behind "combinational" name is that combinational gates have no memory and their reset functions are inversions of their set functions. However, a more typical/easier-to-understand name would be `complexGate`.

Perhaps, we can rename this one into `combinationalComplexGate` and add `complexGate` for set & reset?

jrbeaumont: I like this idea. The name of the function is becoming long, but it is good. We could do something like `complexGate` for just a `set` function, and `complexSRGate` for `set` and `reset`? Less understandable, I'll admit.

snowleopard:
@jrbeaumont Nice tests! I've added a few comments.

Can you also examples to `Tuura.Boolean` functions? Similar to how I do it here: 

http://hackage.haskell.org/package/algebraic-graphs-0.0.5/docs/Algebra-Graph.html#v:hasVertex

2-3 examples per function would be great. Also, make sure you are using Haddock comments syntax, so you can then run `stack haddock` and get nicely formatted HTML documentation.

The examples you add should also be added into tests, to make sure they are correct. For example, here are tests for the above documentation example: 

https://github.com/snowleopard/alga/blob/master/test/Algebra/Graph/Test/Generic.hs#L395-L405

src/Tuura/Concept/Circuit/Derived.hs
snowleopard: Yes, I also thought about something like `complexSRGate`, but I think a longer but clearer name is better. Users can always introduce shorter aliases if they like.

jrbeaumont: We could change this to: 
```haskell
combinationalComplexGate :: Eq a => Expr a -> Transition a -> CircuitConcept a
```
And allow a user to pass `rise x` or `fall x` into this, so they are effectively setting `set`/`reset` themselves. If they do 
```haskell 
combinationalComplexGate (And (Var a) (Var b)) (rise x) <> combinationalComplexGate (Or (Not (Var a)) (Not (Var b))) (fall x)
```
As an example, and then also implement `complexGate` with a set and reset function.

snowleopard: Aha, nice idea!

Perhaps, we could call it simply `function :: Eq a => Expr a -> Transition a -> CircuitConcept a` and also define `complexGate :: Eq a => Expr a -> a -> CircuitConcept a` as a composition of set and reset functions.

jrbeaumont: Would `complexGate` not need to be `Eq a => Expr a -> Expr a -> a -> CircuitConcept`

jrbeaumont: Or will `complexGate` be for only when a set function is provided?

snowleopard: Yes, you are right, we need two `Expr`s.

jrbeaumont: Cool, I'll get on that. And your comment suggestions. Haddock comments will be very useful. Might take me a little while.

snowleopard: Ideally I'd like that `complexGate (And a b) (And (Not a) (Not b)) c` is the same as `cElement a b c` :)

snowleopard: And `function (And a b) (rise c) <> function (Or a b) (fall c)` is the same as `andGate a b c`.

jrbeaumont: I can work that one out, and even throw it in some testing :)

jrbeaumont: By the way, it would need to be `Not (Var a)` etc. I can change this to `Not a == Not (Var a)` if you would like?

snowleopard: 👍

snowleopard: Yes, you are right about `Var`s, we do need them.

snowleopard:
@jrbeaumont By the way: in order to not blow up this PR too much, why don't you fix the minor things I mentioned in comments, leaving Haddock and examples for another PR?

jrbeaumont:
A good idea. Let's get this merged, and I can do Haddock etc for the whole repo in another PR.

snowleopard:
Good, let me know once the minor things are fixed.

src/Tuura/Concept/Circuit/Derived.hs
jrbeaumont:
```haskell
complexGate (And (Var a) (Var b)) (And (Not (Var a)) (Not (Var b))) c
```
Correctly produces a C-element :)

snowleopard: Awesome :)

snowleopard: Btw, you can also write it as follows:

``complexGate (Var a `And` Var b) (Not (Var a) `And` Not (Var b)) c``

This is a bit more readable.

jrbeaumont:
```haskell
function ((Var a) `And` (Var b)) (rise c) <> function ((Not (Var a)) `Or` (Not (Var b))) (fall c)
```
Produces an AND gate :)

jrbeaumont:
@snowleopard: I have fixed those issues. Please check my implementation of `function` and `complexGate` before merging.

src/Tuura/Concept/Circuit/Derived.hs
```diff
@@ -202,13 +202,13 @@ internals ints = interfaceConcept $ \s ->
...
-      toTransitions = map (map (\l -> Transition (variable l) (polarity l)))
+function :: Eq a => Expr a -> Transition a -> CircuitConcept a
+function cause effect = mconcat $ map (toConcept) (toTransitions cnf)
+  where
+    toConcept c = c ~|~> effect
+    cnf = fromCNF $ simplifyCNF $ convertToCNF cause
+    toTransitions = map (map (\l -> Transition (variable l) (polarity l)))
+
+complexGate :: Eq a => Expr a -> Expr a -> a -> CircuitConcept a
+complexGate set reset sig = function set (Transition sig True)
```
snowleopard: Should `Transition sig True` be just `rise sig`?

snowleopard:
Just one last tweak and I'll merge :)

src/Tuura/Concept/Circuit/Derived.hs
jrbeaumont: Of course. Sorry, just going too far on it haha!

jrbeaumont:
@snowleopard There we go!

snowleopard:
Great, thanks! This is now merged.

PR #81, opened 2017-08-08
boolean: show instances for Literal, CNF, DNF. Self-dual functions.

ARandomOWL:
The show instances are useful when manipulating boolean equations directly.

### DIFF
--- a/src/Tuura/Boolean.hs
+++ b/src/Tuura/Boolean.hs
@@ -1,3 +1,7 @@
+{-# LANGUAGE FlexibleInstances #-}
+--{-# LANGUAGE UndecidableInstances #-}
+--{-# LANGUAGE TypeSynonymInstances #-}
+
 module Tuura.Boolean (
   module Tuura.Boolean.Parser,
   CNF (..), DNF (..), Literal (..),
@@ -15,9 +19,22 @@ import Tuura.Concept.Circuit
 
 newtype CNF a = CNF { fromCNF :: [[Literal a]] }
 
+instance Show (CNF String) where
+  show = eqShow "*" " + " ((flip (++) ")") . (++) "(") fromCNF
+
 newtype DNF a = DNF { fromDNF :: [[Literal a]] }
 
-data Literal a = Literal { variable :: a, polarity :: Bool } deriving (Eq, Ord, Show)
+instance Show (DNF String) where
+  show = eqShow " + " "*" id fromDNF
+
+eqShow i1 i2 f1 f2 x = intercalate i1 $ map (f1 . intercalate i2 . map (tidyString . show)) (f2 x)
+  where tidyString = filter (not . (`elem` "\""))
+
+data Literal a = Literal { variable :: a, polarity :: Bool } deriving (Eq, Ord)
+
+instance Show (Literal String) where
+  show (Literal var True) = show var
+  show (Literal var False) = show ("!" ++ var)
 
 convertToCNF :: Eq a => Expr a -> CNF a
 convertToCNF expr = cnf
--- a/src/Tuura/Boolean.hs
+++ b/src/Tuura/Boolean.hs
@@ -1,6 +1,4 @@
-{-# LANGUAGE FlexibleInstances #-}
---{-# LANGUAGE UndecidableInstances #-}
---{-# LANGUAGE TypeSynonymInstances #-}
+{-# LANGUAGE FlexibleInstances #-}  -- For Show instance of Literal String
 
 module Tuura.Boolean (
   module Tuura.Boolean.Parser,
--- a/src/Tuura/Boolean.hs
+++ b/src/Tuura/Boolean.hs
@@ -1,9 +1,8 @@
-{-# LANGUAGE FlexibleInstances, OverlappingInstances, IncoherentInstances #-}
+{-# LANGUAGE FlexibleInstances #-}
 
 module Tuura.Boolean (
   module Tuura.Boolean.Parser,
   CNF (..), DNF (..), Literal (..),
-  prettyCNF, prettyDNF,
   convertToCNF, genConcepts,
   simplifyCNF, simplifyDNF, convertCNFtoDNF) where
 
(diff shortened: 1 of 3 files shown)


src/Tuura/Boolean.hs
```diff
@@ -1,3 +1,5 @@
+{-# LANGUAGE FlexibleInstances #-}  -- For Show instance of Literal String
```
snowleopard: Let's drop the comment, it'd be hard to maintain it.

src/Tuura/Boolean.hs
```diff
@@ -15,9 +17,22 @@ import Tuura.Concept.Circuit
 
 newtype CNF a = CNF { fromCNF :: [[Literal a]] }
 
+instance Show (CNF String) where
+  show = eqShow "*" " + " ((flip (++) ")") . (++) "(") fromCNF
```
snowleopard: It is often assumed that `show` results can be compiled. I would recommend to use derived `Show` instance here, but if you need a nicer representation for readability, you can call this function `pretty`, which seems to be a standard convention.

snowleopard: Same for other `Show` instances below.

src/Tuura/SelfDual.hs
```diff
@@ -0,0 +1,56 @@
+module Tuura.SelfDual (
+    isSelfDual, getSelfDuals) where
+
+import Data.List (sort)
+import Control.Monad (replicateM)
+
+import Tuura.Boolean
```
snowleopard: Aha, it looks like `Tuura.Boolean` deserves to be a separate library, so everyone can reuse it without changing Plato :-)

src/Tuura/SelfDual.hs
```diff
@@ -0,0 +1,56 @@
...
+
+isSelfDual :: String -> Bool
```
snowleopard: Let's separate two concerns: (1) parsing a function from a `String` and (2) testing that the function is self-dual. It would make (2) easier to reuse.

src/Tuura/SelfDual.hs
```diff
@@ -0,0 +1,56 @@
...
+          top = take halfCells
+          bot = reverse . drop halfCells
+          testMutex x = (not . or) (zipWith (&&) (top x) (bot x))
+          good = [ x | x <- possibles, count x == halfCells, testMutex x]
+          count = length . filter (==False)
+
+parseToCNF :: String -> CNF String
+parseToCNF = simplifyCNF . convertToCNF . right . parseExpr
+    where right (Right x) = x
+          right (Left _) = right (parseExpr "")
```
snowleopard: Not sure what happens here: why not report an error instead of returning a meaningless empty `CNF`?

src/Tuura/SelfDual.hs
```diff
@@ -0,0 +1,56 @@
...
+showBool True = '1'
+showBool False = '0'
+
+toBin :: Int -> [Int]
+toBin 0 = [0]
+toBin n = reverse (helper n)
+    where helper 0 = []
+          helper n = let (q,r) = n `divMod` 2 in r : helper q
+
+intToBool :: Int -> Bool
```
snowleopard: I think this is the standard function `toEnum`?

src/Tuura/SelfDual.hs
```diff
@@ -0,0 +1,56 @@
...
+    where right (Right x) = x
+          right (Left _) = right (parseExpr "")
+
+parseToDNF :: String -> DNF String
+parseToDNF = simplifyDNF . convertCNFtoDNF . simplifyCNF . parseToCNF
+
+dualDNF :: DNF a -> CNF a
+dualDNF = CNF . fromDNF
+
+showBool :: Bool -> Char
```
snowleopard: This appears to be unused code?

src/Tuura/SelfDual.hs
```diff
@@ -0,0 +1,56 @@
...
+showBool True = '1'
+showBool False = '0'
+
+toBin :: Int -> [Int]
```
snowleopard: This is also unused?

src/Tuura/SelfDual.hs
snowleopard: Also, `parseToCNF` appears to be unused.

src/Tuura/SelfDual.hs
snowleopard: And this is also unused.

snowleopard:
@ARandomOWL Thanks! I've left a bunch of comments above.

src/Tuura/SelfDual.hs
ARandomOWL: Yes, I think this is a good idea!

src/Tuura/SelfDual.hs
ARandomOWL: `showBool` is useful for printing logic vectors. For example one can use `(map . map) showBool (getSelfDuals 3)` for better readability. Maybe it should be an export.

snowleopard: Yes, sure, it should then be exported. Note though that there is `intToDigit`, which allows you to do `intToDigit. fromEnum` instead. I'd say `showBool` is easy enough to implement in the user code and shouldn't be part of the API.

src/Tuura/Boolean.hs
ARandomOWL: I am not clear on the implementation of `pretty`. Do you have an example?

snowleopard: The implementation stays the same: 
```haskell
pretty :: CNF String -> String
pretty = eqShow "*" " + " ((flip (++) ")") . (++) "(") fromCNF
```

This will not violate the usual expectation that `show` produces valid Haskell (it's not always followed, but it's a useful convention).

ARandomOWL:
I have addressed most comments. I will look into dealing with the parsing errors, I'm not too clear on it at the moment.

ARandomOWL:
@snowleopard Parse errors are now handled

src/Tuura/Boolean.hs
```diff
@@ -1,6 +1,9 @@
+{-# LANGUAGE FlexibleInstances, OverlappingInstances, IncoherentInstances #-}
```
snowleopard: Why do you need `OverlappingInstances` and `IncoherentInstances`?

snowleopard: I guess you don't need them, so please remove.

src/Tuura/SelfDual.hs
```diff
@@ -0,0 +1,43 @@
+module Tuura.SelfDual (
+    isSelfDual, getSelfDuals) where
+
+import Control.Monad (replicateM)
+import Data.Either (isRight)
+import Data.List (sort)
+
+import Tuura.Boolean
+
+isSelfDual :: DNF String -> Bool
```
snowleopard: It seems this can be generalised to `Ord a => DNF a -> Bool`, which would be a more useful function.

src/Tuura/Boolean.hs
ARandomOWL: I need them for the `Show` instances of `Literal a`. I want different instances for `String` and everything else (see lines 34--40). Can you suggest a better way?

ARandomOWL: Perhaps a `prettyLiteral` function like for the CNF and DNF types is what is needed.

snowleopard: Yes, your current `Show` instance does not produce output that can be compiled. I suggest to use `derived Show` instance instead, and add a `prettyLiteral :: Literal String -> String` function.

By the way, if you'd like to avoid using names like `prettyCNF` and `prettyLiteral`, and have a single function instead, have a look at the standard `pretty` library. Specifically, there is a `Pretty` typeclass, which comes with `prettyShow` function that seems to do what you want:

https://hackage.haskell.org/package/pretty-1.1.3.5/docs/Text-PrettyPrint-Annotated-HughesPJClass.html

Feel free to add `Pretty` instances for `CNF` and `Literal`.

ARandomOWL: Thanks, that's exactly what I was looking for. I will add this soon.

ARandomOWL:
@snowleopard I have changed over to the PrettyPrint library. I'm still not happy with the code. Problems stem from the `pPrint String` instance being implemented as `text . show`. I would be grateful of suggestions for improvement.

snowleopard:
@ARandomOWL What about defining only the `Pretty (Literal String)` instance? If you drop the more general `Pretty a => Pretty (Literal a)` there will be no overlapping. Users can then define `Pretty` instances for literals of their favourite types themselves. Same for `DNF`/`CNF`.

ARandomOWL:
@snowleopard I agree it makes sense to remove the generic instances for now. I have pushed that but Appveyor seems to be having problems unrelated to my code.

src/Tuura/SelfDual.hs
```diff
@@ -0,0 +1,43 @@
...
+          possibles = replicateM cells [True,False]
+          top = take halfCells
+          bot = reverse . drop halfCells
+          testMutex x = (not . or) (zipWith (&&) (top x) (bot x))
+          good = [ x | x <- possibles, count x == halfCells, testMutex x]
+          count = length . filter (==False)
+
+parseToCNF :: String -> CNF String
+parseToCNF func =
+    if isRight parsed
```
snowleopard: You can use pattern-matching here to avoid partial functions like `right`:
```haskell
case parseExpr func of
    Left _  -> error $ "..."
    Right x -> ...
```

snowleopard:
OK, looks good -- just one minor comment.

src/Tuura/SelfDual.hs
ARandomOWL: That looks much better. Thanks for the suggestion, it's incorporated now.

ARandomOWL:
I've addressed the comment, and now the checks pass :)

snowleopard:
👍

PR #82, opened 2017-08-11
Added generalised multi-input gates.

jrbeaumont:
`cElementN` for multi-input C-element.
`orGateN` for multi-input OR gate.
`andGateN` for multi-input AND gate.

Also changed some signal names to match with conventions, mainly using `z` for outputs, and so it
matches the thesis.

### DIFF
--- a/src/Tuura/Concept/Circuit/Derived.hs
+++ b/src/Tuura/Concept/Circuit/Derived.hs
@@ -7,7 +7,7 @@ module Tuura.Concept.Circuit.Derived (
     initialise0, initialise1,
     (~>), (~|~>), (~&~>),
     buffer, inverter, cElement, meElement,
-    andGate, orGate, xorGate, srLatch,
+    andGate, orGate, xorGate, srLatch, srLatch2,
     mutex, never, handshake,
     handshake00, handshake11,
     cElementN, orGateN, andGateN,
@@ -172,7 +172,10 @@ xorGate i1 i2 o = [rise i1, rise i2] ~|~> rise o <> [fall i1, fall i2] ~|~> rise
                <> [rise i1, fall i2] ~|~> fall o <> [fall i1, rise i2] ~|~> fall o
 
 srLatch :: Eq a => a -> a -> a -> CircuitConcept a
-srLatch s r q = complexGate ((Var s) `And` (Not (Var r))) ((Var r) `And` (Not (Var s))) q
+srLatch s r q = complexGate (Var s) (Var r) q
+
+srLatch2 :: Eq a => a -> a -> a -> a -> CircuitConcept a
+srLatch2 s r q nq = srLatch s r q <> bubble nq (srLatch s r nq)
 
 -- Protocol-level concepts
 handshake :: a -> a -> CircuitConcept a
--- a/src/Tuura/Concept/Circuit/Derived.hs
+++ b/src/Tuura/Concept/Circuit/Derived.hs
@@ -7,7 +7,7 @@ module Tuura.Concept.Circuit.Derived (
     initialise0, initialise1,
     (~>), (~|~>), (~&~>),
     buffer, inverter, cElement, meElement,
-    andGate, orGate, xorGate, srLatch, srLatch2,
+    andGate, orGate, xorGate,
     mutex, never, handshake,
     handshake00, handshake11,
     cElementN, orGateN, andGateN,
@@ -170,12 +170,6 @@ andGate i1 i2 o = dual $ orGate i1 i2 o
 xorGate :: a -> a -> a -> CircuitConcept a
 xorGate i1 i2 o = [rise i1, rise i2] ~|~> rise o <> [fall i1, fall i2] ~|~> rise o
                <> [rise i1, fall i2] ~|~> fall o <> [fall i1, rise i2] ~|~> fall o
-
-srLatch :: Eq a => a -> a -> a -> CircuitConcept a
-srLatch s r q = complexGate (Var s) (Var r) q <> never [rise s, rise r]
-
-srLatch2 :: Eq a => a -> a -> a -> a -> CircuitConcept a
-srLatch2 s r q nq = srLatch s r q <> bubble nq (srLatch s r nq) <> mutex q nq
 
 -- Protocol-level concepts
 handshake :: a -> a -> CircuitConcept a
--- a/src/Tuura/Concept/Circuit/Derived.hs
+++ b/src/Tuura/Concept/Circuit/Derived.hs
@@ -172,10 +172,10 @@ xorGate i1 i2 o = [rise i1, rise i2] ~|~> rise o <> [fall i1, fall i2] ~|~> rise
                <> [rise i1, fall i2] ~|~> fall o <> [fall i1, rise i2] ~|~> fall o
 
 srLatch :: Eq a => a -> a -> a -> CircuitConcept a
-srLatch s r q = complexGate (Var s) (Var r) q
+srLatch s r q = complexGate (Var s) (Var r) q <> mutex s r
 
 srLatch2 :: Eq a => a -> a -> a -> a -> CircuitConcept a
-srLatch2 s r q nq = srLatch s r q <> bubble nq (srLatch s r nq)
+srLatch2 s r q nq = srLatch s r q <> bubble nq (srLatch s r nq) <> mutex q nq
 
 -- Protocol-level concepts
 handshake :: a -> a -> CircuitConcept a
--- a/src/Tuura/Concept/Circuit/Derived.hs
+++ b/src/Tuura/Concept/Circuit/Derived.hs
@@ -172,7 +172,7 @@ xorGate i1 i2 o = [rise i1, rise i2] ~|~> rise o <> [fall i1, fall i2] ~|~> rise
                <> [rise i1, fall i2] ~|~> fall o <> [fall i1, rise i2] ~|~> fall o
 
 srLatch :: Eq a => a -> a -> a -> CircuitConcept a
-srLatch s r q = complexGate (Var s) (Var r) q <> mutex s r
+srLatch s r q = complexGate (Var s) (Var r) q <> never [rise s, rise r]
 
 srLatch2 :: Eq a => a -> a -> a -> a -> CircuitConcept a
 srLatch2 s r q nq = srLatch s r q <> bubble nq (srLatch s r nq) <> mutex q nq
--- a/src/Tuura/Concept/Circuit/Derived.hs
+++ b/src/Tuura/Concept/Circuit/Derived.hs
@@ -9,6 +9,7 @@ module Tuura.Concept.Circuit.Derived (
     buffer, inverter, cElement, mutexElement,
     andGate, orGate, xorGate, mutex, never, handshake,
     handshake00, handshake11,
+    cElementN, orGateN, andGateN,
     inputs, outputs, internals, function, complexGate
     ) where
 
@@ -184,6 +185,16 @@ mutex a b = fall a ~> rise b <> fall b ~> rise a <> never [rise a, rise b]
 
 never :: [Transition a] -> CircuitConcept a
 never es = invariantConcept (NeverAll es)
+
+-- Generalized multi-input gates
+cElementN :: [a] -> a -> CircuitConcept a
+cElementN ins out = mconcat $ map (`buffer` out) ins
+
+orGateN :: [a] -> a -> CircuitConcept a
+orGateN ins out = map rise ins ~|~> rise out <> map fall ins ~&~> fall out
+
+andGateN :: [a] -> a -> CircuitConcept a
+andGateN ins out = dual $ orGateN ins out
 
 -- Signal type declaration concepts
 inputs :: Eq a => [a] -> CircuitConcept a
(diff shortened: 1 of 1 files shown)


src/Tuura/Concept/Circuit/Derived.hs
```diff
@@ -147,28 +148,27 @@ initialise1 as = initialise (head as) True <> initialise1 (tail as)
 
 -- Gate-level concepts
 buffer :: a -> a -> CircuitConcept a
-buffer a b = rise a ~> rise b <> fall a ~> fall b
+buffer a z = rise a ~> rise z <> fall a ~> fall z
```
snowleopard: The reader of this code will not necessarily be familiar with the convention to use `a...` for inputs and `x...` for outputs. If you'd like to make the code easier to understand, I think it would be better to use `buffer i o`, `cElement i1 i2 o`, etc.

src/Tuura/Concept/Circuit/Derived.hs
```diff
@@ -180,12 +180,22 @@ handshake00 a b = handshake a b <> initialise a False <> initialise b False
...
 
 never :: [Transition a] -> CircuitConcept a
 never es = invariantConcept (NeverAll es)
 
+-- Generalized multi-input gates
+cElementN :: [a] -> a -> CircuitConcept a
+cElementN ins out = mconcat $ map (`buffer` out) ins
+
+orGateN :: [a] -> a -> CircuitConcept a
+orGateN ins out = map rise ins ~|~> rise out <> map fall ins ~&~> fall out
```
snowleopard, line 195: Oh, that's nice!

src/Tuura/Concept/Circuit/Derived.hs
```diff
@@ -147,28 +148,27 @@ initialise1 as = initialise (head as) True <> initialise1 (tail as)
...
-cElement a b c = buffer a c <> buffer b c
+cElement a b z = buffer a z <> buffer b z
 
-meElement :: a -> a -> a -> a -> CircuitConcept a
-meElement r1 r2 g1 g2 = buffer r1 g1 <> buffer r2 g2 <> me g1 g2
-
-andGate :: a -> a -> a -> CircuitConcept a
-andGate a b c = rise a ~> rise c <> rise b ~> rise c
-             <> [fall a, fall b] ~|~> fall c
+mutexElement :: a -> a -> a -> a -> CircuitConcept a
```
snowleopard: I think 'ME element' is a more common name for this gate, let's keep it. The change of `me` to `mutex` is fine though.

snowleopard:
@jrbeaumont Thanks! I've left a few minor comments.

By the way, can you also define `srLatch` via the `complexGate`?

This brings me to another interesting higher-order transformation that can perhaps be called `enable`. It turns any concept into a concept that has additional control input `enable`. With this you could then define `srLatchWithEnable s r e = enable (rise e) (srLatch s r)`, which seems pretty cool I think :-) But let's do this in a different PR. See https://en.wikipedia.org/wiki/Flip-flop_(electronics)#Gated_SR_latch.

src/Tuura/Concept/Circuit/Derived.hs
```diff
@@ -148,27 +149,30 @@ initialise1 as = initialise (head as) True <> initialise1 (tail as)
...
+andGate i1 i2 o = dual $ orGate i1 i2 o
 
 xorGate :: a -> a -> a -> CircuitConcept a
-xorGate a b z = [rise a, rise b] ~|~> rise z <> [fall a, fall b] ~|~> rise z
-             <> [rise a, fall b] ~|~> fall z <> [fall a, rise b] ~|~> fall z
+xorGate i1 i2 o = [rise i1, rise i2] ~|~> rise o <> [fall i1, fall i2] ~|~> rise o
+               <> [rise i1, fall i2] ~|~> fall o <> [fall i1, rise i2] ~|~> fall o
+
+srLatch :: Eq a => a -> a -> a -> CircuitConcept a
+srLatch s r q = complexGate ((Var s) `And` (Not (Var r))) ((Var r) `And` (Not (Var s))) q
```
jrbeaumont: @snowleopard: what do you think of `srLatch`?

snowleopard: Two comments:
* The classic definition is simpler `complexGate (Var s) (Var r) q`.
* We could also have a version with two outputs `q` and `nq`? Not sure about the best name, but perhaps `srLatch2 s r q nq = srLatch s r q <> dual (srLatch s r nq)` :)

src/Tuura/Concept/Circuit/Derived.hs
```diff
@@ -172,7 +172,10 @@ xorGate i1 i2 o = [rise i1, rise i2] ~|~> rise o <> [fall i1, fall i2] ~|~> rise
                <> [rise i1, fall i2] ~|~> fall o <> [fall i1, rise i2] ~|~> fall o
 
 srLatch :: Eq a => a -> a -> a -> CircuitConcept a
-srLatch s r q = complexGate ((Var s) `And` (Not (Var r))) ((Var r) `And` (Not (Var s))) q
+srLatch s r q = complexGate (Var s) (Var r) q
+
+srLatch2 :: Eq a => a -> a -> a -> a -> CircuitConcept a
+srLatch2 s r q nq = srLatch s r q <> bubble nq (srLatch s r nq)
```
jrbeaumont: @snowleopard `dual` didn't quite work as well for `nq`. But `bubble bq` is the exact operation we desire.

snowleopard:
I think this requires a bit more discussion.

Firstly, for `q` we have `complexGate (Var s) (Var r) q`, i.e. `rise s ~> rise q <> rise r ~> fall q`. Furthermore,  the input combination `s=r=1` is usually forbidden, which can be stated by using the `never` concept. Note that this is a constraint on inputs (not outputs like in the ME element), so if you leave `s` and `r` unrestricted, the invariant verification should fail. You can fix this by composing this with `mutex s r` constraint in the environment of the SR latch. It's interesting and worth discussing in the thesis.

For `nq` we'd like to have `complexGate (Var r) (Var s) nq`, i.e. `rise r ~> rise nq <> rise s ~> fall nq`. Similarly, this needs a `never` constraint.

Now it looks like we can prove that you are right and we need `bubble nq` instead of `dual`. Indeed, if we compute `bubble nq $ rise s ~> rise nq <> rise r ~> fall nq`, we get the above formula for `nq`! Moreover the `never` constraint appears to be unaffected, as desired. And the initial state becomes is inverted. Very cool and not straightforward at all -- this should give you 2-3 pages for the thesis :-)

Can you verify that you can synthesise a usual implementation from this `srLatch2` spec? This might require some fiddling with MPSAT command line flags to allow to only use 2-input NOR gates.

snowleopard:
By the way, maybe `srLatch2` needs another output constraint: `mutex q nq`.

jrbeaumont:
I thought it was necessary to include a `mutex` for `s` and `r`, but I worried it would be trying to constrain the environment, which if we just use `never [rise s, rise r]` then this would be problematic, because without the `mutex`, we cannot force one to be low for the other to go high. But you are right, we can add it nicely.
I will add `mutex s r` to the `srLatch`, which should also apply to `srLatch2`. I will also add `mutex q nq` to `srLatch2`.

This does provide a lot to discuss in the thesis. So far, the gates we have are reasonably straight forward, except maybe the XOR, but XOR requires further discussion about how the set/reset functions are used, and why CNF is required in concepts, and so on. Why `srLatch` and `srLatch2` are implemented as they are. However, I wonder if I may need to discuss `srLatch2` as part of the `bubble` section, as it does rely quite heavily on `bubble`, which would be tought to explain before introducing it formally.

jrbeaumont:
@snowleopard I have added the mutexes. In result to your question:
> Can you verify that you can synthesise a usual implementation from this srLatch2 spec? This 
> might require some fiddling with MPSAT command line flags to allow to only use 2-input NOR 
> gates.
Yes and No. The resulting STG of an `srLatch2` is:
image

The result of synthesis, using `-g2` for only 2 input gates:

image

Ultimately, it works as expected, but the implementation is non-optimal.

src/Tuura/Concept/Circuit/Derived.hs
```diff
@@ -172,10 +172,10 @@ xorGate i1 i2 o = [rise i1, rise i2] ~|~> rise o <> [fall i1, fall i2] ~|~> rise
                <> [rise i1, fall i2] ~|~> fall o <> [fall i1, rise i2] ~|~> fall o
 
 srLatch :: Eq a => a -> a -> a -> CircuitConcept a
-srLatch s r q = complexGate (Var s) (Var r) q
+srLatch s r q = complexGate (Var s) (Var r) q <> mutex s r
```
snowleopard: Wait, I think `mutex` is too restrictive -- in this way you restrict the behaviour of inputs. You need `never` here, because this is just an assumption about the environment, but not a restriction. In this way, the verification tools will spot an error if inputs are not mutually exclusive.

snowleopard:
> However, I wonder if I may need to discuss `srLatch2` as part of the `bubble` section, as it does rely quite heavily on `bubble`, which would be tought to explain before introducing it formally.

I think it is worth studying `srLatch` and friends as a final example of higher-order transformations, because they allow you to highlight both `bubble` and `enable`. By the way, what is the dual of `srLatch`?

snowleopard:
> Ultimately, it works as expected, but the implementation is non-optimal.

@jrbeaumont Hmm, this needs further investigation, I think.

Can you come up with an STG that leads to the circuit we expect? We'll then be able to see where the mismatch is coming from.

src/Tuura/Concept/Circuit/Derived.hs
jrbeaumont: Ahh, right. As I feared. On import to Workcraft of just an `srLatch` or `srLatch2`, this will state that the `never` state can be reached, but will not stop the import.

jrbeaumont:
>because they allow you to highlight both bubble and enable. By the way, what is the dual of srLatch?

I will add it to the end of the transformations section then, as examples, but also state they are included in the library. I will check the dual right away, and also find out where the mismatch comes from!

src/Tuura/Concept/Circuit/Derived.hs
snowleopard: Yes, that's the correct behaviour!

If you'd like to actually restrict inputs for the purposes of testing in the right environment, you can also append `me s r`. This will then restrict the inputs, perhaps giving more opportunities for synthesis tools.

snowleopard: Just to clarify what I mean, you can import for following spec:
```
circuit s r q = srLatch s r q <> me s r
```

jrbeaumont: Aha yes. I have been doing this with testing :)

jrbeaumont:
OK, I have found out what the problem is. 
Because we use two `srLatch`es, the output signals are both based on both of the inputs. 
Whereas, if have:
```haskell 
example s r q nq = rise r ~> fall q <> fall r ~> rise q 
                <> rise s ~> fall nq <> fall s ~> rise nq 
                <> mutex q nq
```
Then the mutex of `q` and `nq` ensures that when one signal rises, the other falls, but the inputs cause only one of them to change. They are inverters, because when `s` is set, before `q` can rise, `nq` must fall, requiring `s` to cause `nq` to fall, and vice versa.

jrbeaumont:
Also, is the dual operation an meElement?

snowleopard:
@jrbeaumont Aha, I think you're right, there is a close relation between `srLatch2` and `meElement`. Looks like `srLatch2 s r q nq = never [rise s, rise r] <> bubble s (bubble r (meElement r s q nq))`.

Basically, we swap and invert the inputs, and also add an assumption that inputs are never high at the same time. The latter is equivalent to forbidding `r1=r2=0` combination for the ME element, which means there will never be any metastability (there are no races between `r1+` and `r2+`). Makes sense! 

Do you agree with this?

I propose to take care of `srLatch` and friends in a separate PR -- probably together with `enable`, as it turned out to be a bit more complicated than I anticipated.

P.S.: When writing `bubble s (bubble r (...` above I realised that the `bubble` transformation is commutative. That is, `bubble x . bubble y == bubble y . bubble x`. Can you point this out in the thesis?

jrbeaumont:
@snowleopard I noticed a similarity, but wasn't quite sure. But this does make sense, I agree :)
It is demonstrable in how we implement the meElement, using buffers and such. Having `r` inverted to provide `q` and `s` inverted to provide `nq` is like having `buffer r1 g1` and `buffer r2 g2`. In the case of an srLatch, we use inverters instead, as with the `bubbles`, and swap them so that, r1 and g2 match up etc, but the mutex between g1 and g2 holds.  The `never [rise s, rise r]` as you say, is just the added expectations of the `srLatch`, you cannot set and reset at the same time.

I hope that makes sense, it is how I interpreted what you said above. 

I will remove all `srLatch`es from this PR, and create a quick issue about including `srLatch`, `srLatch2` and `enable`, and we can discuss things in that issues/the upcoming PR.

>P.S.: When writing bubble s (bubble r (... above I realised that the bubble transformation is commutative. That is, bubble x . bubble y == bubble y . bubble x. Can you point this out in the thesis?

I have already done this in my most recent corrections. I was adding a simple proof, and decided to point out that commutativity applies too :)

jrbeaumont:
`srLatch` etc is removed. Feel free to merge when the CI has passed.

snowleopard:
@jrbeaumont Thanks! Merged.

> I have already done this in my most recent corrections. I was adding a simple proof, and decided to point out that commutativity applies too :)

Great :)

Issue #83, opened 2017-08-15
Implement srLatch and enable.


jrbeaumont:
Enable is another higher order transformation for concepts, which applies a given signal as an `enable` signal to the given concept, 
i.e. the function of the concept will not apply unless the `enable` signal is high (or low).

`srLatch` is a useful concept to include, and we have discussed some details of it in #82.
There are some implementation details which need working out, such as it being based on an `meElement`.

These will need explanation in the thesis too. `enable` can be explained in the details of how it is implemented, and how useful this can be.

`srLatch`es implementation details may be more complex to discuss, such as how it is based off an `meElement` and the transformations applied to it as a cause of this. As such, it would be best to explain this as an example in the Concept Transformation section of the thesis. 

First however, we need to establish the implementation of `enable` and `srLatch`/`srLatch2`

snowleopard:
I'd like to propose to change the naming slightly: `srLatch -> srHalfLatch` and `stLatch2 -> stLatch`, which I think we'll be more convenient, since SR latches pretty often have two outputs.

snowleopard:
With the above name change, I think the following holds: 

`srLatch s r q nq = never [rise s, rise r] <> bubbles [s, r] (meElement r s q nq)`

Here `bubbles` is a straightforward generalisation of `bubble` to lists of signals.

Furthermore, we expect that `srHalfLatch s r q` is equivalent to `internal [nq] <> srLatch s r q nq` in terms of behaviour.

snowleopard:
Some equational reasoning to follow up @jrbeaumont's comments on inverters:
```haskell
srLatch s r q nq
= never [rise s, rise r] <> bubbles [s, r] (meElement r s q nq)
= never [rise s, rise r] <> bubbles [s, r] (buffer r q <> buffer s nq <> mutex q nq)
= never [rise s, rise r] <> inverter r q <> inverter s nq <> mutex q nq
```

jrbeaumont:
>I'd like to propose to change the naming slightly:

I agree with the name changes, they are much clearer.

>`srLatch s r q nq = never [rise s, rise r] <> bubbles [s, r] (meElement r s q nq)`

This is great, a nice implementation, and I like the reasoning in your last comment, this would be good to explain in the thesis.
For `srHalfLatch`, I may need to do some testing and playing around to find the correct implementation, but I beleive `srHalfLatch s r q = rise s ~> rise q <> rise r ~> fall q <> never [rise s, rise r]`, as we have implemented it before would work. 

[...]

PR #84, opened 2017-08-16
Implementations of concepts and transformations discussed in #83

jrbeaumont:


src/Tuura/Concept/Circuit/Derived.hs
```diff
@@ -171,6 +183,19 @@ xorGate :: a -> a -> a -> CircuitConcept a
 xorGate i1 i2 o = [rise i1, rise i2] ~|~> rise o <> [fall i1, fall i2] ~|~> rise o
                <> [rise i1, fall i2] ~|~> fall o <> [fall i1, rise i2] ~|~> fall o
 
+-- Set/reset latch including one non-inverted and one inverted output.
+srLatch :: Eq a => a -> a -> a -> a -> CircuitConcept a
+srLatch s r q nq = never [rise s, rise r]      -- disallow contradictory requests
+    <> initialise0 [q]   <> initialise1 [nq]   -- disallow initial state q=nq=0
```
snowleopard: I think a better comment here would be `set initial state to q=0 and nq=1`. And add the following on top of the function:

`TODO: Generalise by allowing both 01 and 10 as initial states.`

src/Tuura/Concept/Circuit/Derived.hs
```diff
@@ -89,41 +89,53 @@ dual c = mempty
...
 toggleSpecific a t
-    | signal t == a = toggle t
-    | otherwise     = t
+    | signal t `elem` a = toggle t
+    | otherwise         = t
 
 -- For the selected signal, toggle the directions of all transitions for this.
-bubbleCausality :: Eq a => a -> Causality (Transition a)
+bubbleCausality :: Eq a => [a] -> Causality (Transition a)
                         -> Causality (Transition a)
```
snowleopard: Don't break this type signature, it's OK to occasionally exceed the line width limit.

src/Tuura/Concept/Circuit/Derived.hs
```diff
@@ -89,41 +89,53 @@ dual c = mempty
...
 bubbleCausality s (Causality f t)
-    | signal t == s = Causality (map (toggleSpecific s) f) (toggle t)
-    | otherwise     = Causality (map (toggleSpecific s) f) t
+    | signal t `elem` s = Causality (map (toggleSpecific s) f) (toggle t)
+    | otherwise         = Causality (map (toggleSpecific s) f) t
 
 -- Invert the invariant transition of the selected signal.
-bubbleInvariant :: Eq a => a -> Invariant (Transition a)
+bubbleInvariant :: Eq a => [a] -> Invariant (Transition a)
                         -> Invariant (Transition a)
```
snowleopard: Same here, let's not break this line.

snowleopard:
@jrbeaumont I think the current version of `enable` is too strict: not only it restricts the behaviour of outputs, it can also prevent inputs from firing, which is probably undesirable. 

Perhaps, it would be a more useful function if it can be used in a more targeted way, for example by applying it only to specified output (or a set of outputs): `enable e [q, nq] $ srLatch s r q nq`. Imagine applying it to all signals in this case -- this would mean inputs can never fire!

It is similar to `bubble/bubbles` in the sense that it is commutative. But it is idempotent (enabling twice is the same as enabling once), whereas `bubble` is an involution, i.e. it's a self-inverse (inverting an input twice is a no-op).

jrbeaumont:
@snowleopard: I didn't think of this! Thank you for pointing it out. Targeting it to specific signals would be a much better idea, or if possible, aiming it only at output/internal signals. However, this is much more difficult to do, as we cannot guarantee that the interface concepts will be included in the concept which is passed into `enable`.

p.s. idempotent and involution are fantastic words :)

jrbeaumont:
@snowleopard I have made some changes, let me know what you think.

src/Tuura/Concept/Circuit/Derived.hs
```diff
@@ -89,41 +89,52 @@ dual c = mempty
...
+bubbles :: Eq a => [a] -> CircuitConcept a -> CircuitConcept a
+bubbles s c = mempty
+    {
+       initial = bubbleInitialValue s (initial c),
+       arcs = fmap (bubbleCausality s) (arcs c),
+       interface = interface c,
+       invariant = fmap (bubbleInvariant s) (invariant c)
+    }
 
+-- Apply enable signal only for transitions of given signals in the concept
```
snowleopard: I think in practice we almost always will apply `enable` only to a single output, so I'd prefer `enable` to take a single signal only. You can add `enables`, which sounds a bit odd, but is probably OK from the symmetry point of view :)

snowleopard:
Wait. The new `enable t s c` can probably be implemented simply as `c <> (t ~> rise s <> t ~> fall s)`. What do you think?

snowleopard:
Which brings me to another useful variant of `enable`, which we can call `sync`. It can be implemented similarly: `sync x y c = c <> (rise x ~> rise y <> fall x ~> fall y)`.

Simplifying even further: `sync x y c = c <> buffer x y`.

Now you can express `meElement` as `sync r1 g1 $ sync r2 g2 $ mutex g1 g2` :)

It doesn't look like a major step forward, but I think it's a fairly flexible higher-order transformation that can make some specifications easier to understand.

snowleopard:
I realised there is a subtle difference between the current implementation and the one I proposed above.

Currently, `enable t s mempty == mempty`.

In the above proposal, `enable t s mempty = t ~> rise s <> t ~> fall s`.

I'm unsure which one is preferable. 

@jrbeaumont Any thoughts?

jrbeaumont:
An enable for a single output can be done, it makes more sense to have it this way, but I would like to have `enables` like you said, for symmetry. 

>Wait. The new enable t s c can probably be implemented simply as c <> (t ~> rise s <> t ~> fall s).

This makes a lot more sense. An output will have constraints on both rise and fall transitions, so it only makes sense to generalise it, and make it more efficient, and ensure that the `enable` signal applies to both of it's transitions.

Sync seems interesting, I am not sure I quite understand it's use in this example. It seems to just replace buffer, but not use standard `<>` operator. Is it an idea to synchronise signals?

As far as for `mempty`, I am also unsure. I think ideally, if someone is to use `enable` on it's own, it should produce a result, it could be seen as a buffer in some cases, i.e. `s` indicates the polarity of `t`, so it is identifiable whether the circuit is enabled.

jrbeaumont:
I have changed `enable`, and added `enables` and `sync`

jrbeaumont:
The clearer `srLatch` implementation has been added.

snowleopard:
@jrbeaumont Thanks, looks good!

PR #87, opened 2017-08-18
Fixed an error in srHalfLatch

jrbeaumont:


snowleopard:
Good catch. There is also a typo in the comment in the very same line! :)

jrbeaumont:
What typo? ;)
(Now all tested and working correctly. Travis seems to be having an issue with macOS today)

snowleopard:
Merged! =)

PR #89, opened 2017-08-24
Migrate Tuura.Boolean to new package tuura-boolean

ARandomOWL:
- genConcepts is moved into Tuura.Concept.Circuit
- Plato now depends on the tuura/boolean commit specified in stack.yaml


jrbeaumont:
I like it :)

snowleopard:
@ARandomOWL Thanks! Looks good :)

PR #93, opened 2017-10-25
Added support for declared signals, and translation of them

jrbeaumont:
This includes:
* Allows concepts with signals as arguments to be translated
* Allows concepts with signals declared in a data type to be translated
* Keyword for concepts with signals as arguments changed to `component` to reflect their usage
* Keyword for concepts with declared signals set as `system`
* Renamed `translate/Main.hs` to `translate/ComponentMain.hs` as this is used only to translate `component` concepts.
* Created `translate/SystemMain.hs` as this is used only to translate `system` concepts.
* Added `convert` function to convert `system` concepts to a form usable by translation.


jrbeaumont:
Also note, `examples/Celement_with_env_1.hs` has been changed to have declared signals for testing purposes.

examples/Celement_with_env_1.hs
```diff
@@ -3,16 +3,19 @@ module Concept where
 import Tuura.Concept.STG
 
 -- C-element with environment circuit described using signal-level concepts
-circuit a b c = interface <> outputRise <> inputFall <> outputFall <> inputRise <> initialState
+
+data Sign = A | B | C deriving (Eq, Enum, Bounded)
```
snowleopard: I'd prefer to keep the full name `Signal` to avoid confusion ("sign" is a word). It's a local name, so there shouldn't be any issues with the name clashes.

examples/Celement_with_env_1.hs
```diff
@@ -3,16 +3,19 @@ module Concept where
...
+    inputFall = rise C ~> fall A <> rise C ~> fall B
 
-    outputFall = fall a ~> fall c <> fall b ~> fall c
+    outputFall = fall A ~> fall C <> fall B ~> fall C
 
-    inputRise = fall c ~> rise a <> fall c ~> rise b
+    inputRise = fall C ~> rise A <> fall C ~> rise B
 
-    initialState = initialise a False <> initialise b False <> initialise c False
+    initialState = initialise A False <> initialise B False <> initialise C False
```
snowleopard: Missing newline?

examples/Celement_with_env_1.hs
```diff
@@ -3,16 +3,19 @@ module Concept where
 import Tuura.Concept.STG
 
 -- C-element with environment circuit described using signal-level concepts
-circuit a b c = interface <> outputRise <> inputFall <> outputFall <> inputRise <> initialState
+
+data Sign = A | B | C deriving (Eq, Enum, Bounded)
+
+circuit = interface <> outputRise <> inputFall <> outputFall <> inputRise <> initialState
```
snowleopard: Should this be named `system` instead?

examples/Celement_with_env_1.hs
jrbeaumont: This was just from testing, the name doesn't affect anything, but I will set it to `Signal`.

examples/Celement_with_env_1.hs
jrbeaumont: Yes, will add back.

src/Tuura/Plato/Translate/Translation.hs
```diff
@@ -93,3 +93,27 @@ toTransitions = map (\l -> Transition (variable l) (polarity l))
...
+convert c = mempty
+         {
+           initial = convertInitialValue (initial c),
+           arcs = fmap convertCausality (arcs c),
+           interface = convertInterface (interface c),
+           invariant = fmap convertInvariant (invariant c)
+         }
+
+convertInitialValue :: Enum a => (a -> InitialValue) -> (Signal -> InitialValue)
+convertInitialValue f (Signal i) = f $ toEnum i
```
snowleopard: I guess this can be generalised to type `Enum a => (a -> b) -> (Signal -> b)` without changing the implementation? Then you don't need to duplicate code in `convertInterface`.

examples/Celement_with_env_1.hs
jrbeaumont: Yes it also should, reminding me that I haven't set anything to check for `system`

snowleopard:
@jrbeaumont Thanks, looks good, just a few comments.

One more suggestion: let's call the files simply `translate/Component.hs` and `translate/System.hs`, i.e. without `Main`. I think it reads better this way.

jrbeaumont:
@snowleopard: Changes made, if you don't have any further suggestions, let's just check it passes CI before we merge :) 
Thanks.

src/Tuura/Plato/Translate/Translation.hs
```diff
@@ -97,21 +97,18 @@ arcLists xs = [ (f, t) | Causality f t <- xs ]
...
+           initial = convertInitsInterface (initial c),
            arcs = fmap convertCausality (arcs c),
-           interface = convertInterface (interface c),
+           interface = convertInitsInterface (interface c),
            invariant = fmap convertInvariant (invariant c)
          }
 
-convertInitialValue :: Enum a => (a -> InitialValue) -> (Signal -> InitialValue)
-convertInitialValue f (Signal i) = f $ toEnum i
+convertInitsInterface :: Enum a => (a -> b) -> (Signal -> b)
```
snowleopard: I think the name is a bit obscure. What about `convertFunction`?

snowleopard:
Thanks! Just one last tweak and I'll merge.

jrbeaumont:
Done :)

snowleopard:
@jrbeaumont Merged. Thank you!

PR #94, opened 2017-12-11
Added combinational gate concept

jrbeaumont:
The `combinationalGate` concept takes in a single Boolean function, the set function of a combinational gate. This type of gate uses this set function for both the set and reset of the gate, simply negating the set function for the reset. Therefore, this concept takes a single function, but produces a set and reset function for the given signal.

### DIFF
--- a/TransTest.hs
+++ b/TransTest.hs
@@ -14,6 +14,7 @@ main = do
     testOrGateConcept
     testTwoAndGates
     testTwoDifferentCElements
+    testBooleanFunctionConcepts
 
 testDotGOutput :: IO ()
 testDotGOutput = do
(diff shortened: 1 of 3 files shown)


snowleopard:
Something off with the CI?

snowleopard (commented):
Looks fine but can you add a test?

jrbeaumont:
The CI issue stems from a problem with not being able to find the stack file once it's downloaded. It should now be fixed.

snowleopard:
@jrbeaumont Yep, your fix seems to work!

Now, let's add a test. For example, something like `combinationalGate (And x y) == andGate x y`.

jrbeaumont:
@snowleopard I have added a test. It tests `combinationalGate` which uses `complexGate`, so this is also tested, and this in turn uses `function` so this is tested too!

snowleopard:
@jrbeaumont Nice! The PR is now merged, thank you.

PR #96, opened 2020-09-28
Update stack.yaml syntax for Stack v1.11

ARandomOWL:


snowleopard (approved):
Looks good to me but perhaps @jrbeaumont would also like to have a look.

jrbeaumont:
Looks good to me too. Thanks for bringing it up to date.

snowleopard:
Thanks @jrbeaumont, merged. I think this is not the full story though, since there is also #97.

@ARandomOWL Thank you for checking up on Plato!

